PR #453, opened 2022-12-04
Updates ioredis to ^5.0.4

janthurau:


jamespantalones:
@janthurau what are the downsides of removing the locks? if each instance sends the same message at the same time, could that cause clientside issues?

we are also looking into forking and dropping locks as we are running on a cluster, which seems to throw errors when a lock attempts to be acquired twice

https://github.com/mike-marcacci/node-redlock#using-clustersentinel

lzj723:
Hi @jamespantalones , you can try below if you just want to solve require lock errors.
It's efficient for me though the code is not that succinct:
`
try {
  // try to get a lock
  const lock = await redlock.acquire(['xxx'], 2000);
  try {
    // do something with lock
  } catch (e) {
    console.error(`log something`, e);
  } finally {
    // Release the lock. And it may throw an ERROR too.
    try {
      await lock.release();
    } catch (e) {
      // console.error('lock release failed:', e);
    }
  }
} catch (e) {
  // It will throw an ERROR when require lock failed, so we need to catch it.
  console.warn('require lock failed:', e);
}
`

janthurau:
I have released v2.1.0-alpha.0 which adds compatibility to ^5.0.4 and removes redlock. Please let me know if that breaks something

PR #483, opened 2023-01-18
docs/cli: Use npx

raineorshine:
npx executes binaries from local packages, which is easier and more future proof than using node_modules paths.

npx is pre-bundled with npm since [v5.2.0](https://github.com/npm/npm/releases/tag/v5.2.0)


janthurau:
First PR after going public :-)

Thanks! 🥳

Issue #455, opened 2022-12-07
Server may not receive Sync message if express-ws handler yields before calling handleConnection


jordanstephens:
**Description**

Following the [Installation docs](https://tiptap.dev/hocuspocus/installation/express#express) for integrating Hocuspocus server with an Express.js app, I found a race condition during the process of responding to a request and setting up the connection.

My express-ws request handler looked like this:

```javascript
app.ws("/path/:id", (ws, req, next) => {
  getContext().then((context) => {
    server.handleConnection(
      ws,
      req,
      req.params.id,
      context
    );
  });
});
```

This resulted in intermittent failure to resolve the connection attempt. Eventually awareness messages would be exchanged updating the HocuspocusProvider's `status` to `Connected`, but the document would be empty.

Eventually, I determined that the HocuspocusProvider's `Sync` message was not being reliably received by the Hocuspocus server. As far as I can tell, this happens because the websocket will emit the `open` event to the HocuspocusProvider before our express-ws route handler gets called. This sets up a race between the provider sending the first `Sync` message and the server setting up listeners to receive it. Success depended on how long it took for my `getContext` function to run.

I noticed that `handleConnection` sets up a queue so it can replay messages received before the listeners are established, but there is still a gap between the `open` event being emitted and `handleConnection` being called. That gap may only appear if the route handler performs async operations and yields control back to the event loop, but it is obviously fragile to depend on avoiding a context switch like this.

In my case, I think I have worked around it by building my own queue in the route handler to replay messages received while `getContext` runs, but perhaps a more robust solution would be to have the server initiate the sync process instead of the Provider. Currently, the Provider is unable to know when it is safe to send the first `Sync` message, so perhaps the first message should be the server indicating its readiness to handle messages.

**Steps to reproduce the bug**
1. Integrate Hocuspocus server into an Express.js app following the [documentation](https://tiptap.dev/hocuspocus/installation/express#express)
2. Sleep in the event loop in the express-ws route handler before calling `server.handleConnection`
3. Observer Hocuspocus provider failing to load document

**Expected behavior**
I expect Hocuspocus server to be ready to receive Sync messages when HocuspocusProvider sends them.

**Environment?**
 - operating system: MacOS 12.6.1
 - browser: Firefox 107 and Chrome 108
 - desktop
 - hocuspocus version: v1.0.0-beta.6

**Additional context**

[...]

Issue #343, opened 2022-05-25
Use 1 WS connection to fetch documents instead of establishing connection for each documents


shahriar-shojib:
Right now hocuspocus provider opens WS connection each time the document needs to change, instead, it should use one connection to handle document changes.

In addition to that, hocuspocus provider should have a plugin system like hocuspocus server where it should be possible to store documents in memory/browser's local storage or implement offline capabilities.

Right now, changing to another document and getting back to the previous document causes the provider to fetch the document from the server before the contents are shown in the editor. Implementing something like this will give us the opportunity to improve the UX of whatever it is we're trying to build.

janthurau:
https://github.com/ueberdosis/hocuspocus/releases/tag/v2.0.0-beta.0

Issue #283, opened 2022-01-14
[RFC] Road to beta


hanspagel:
It’s amazing what we achieved with Hocuspocus already. The adoption is insane, it’s running in production in tons of apps already. BUT, we’re still in alpha, and in a pretty closed state. 👀

We’d like to transition to a more stable beta version soonish. Let’s keep track of what needs to be done in this issue:

- [x] #286
- [x] Release the new Redis extension
- [x] #376
- [x] Migrate all tests to AVA
- [x] Go through the documentation to reflect recent changes and see Hocuspocus from the developers perspective

Things we can do in a beta stage:

- [x] Incorporate feedback for the above changes
- [x] Improve the typing
- [x] Polish the documentation
- [x] Add tons of examples

This list can change in the future. 

Put your feedback/questions below 👇

svenadlung:
I am closing this topic as we did publish the stable version. Please discuss other topics in dedicated issues 🙏.

Issue #249, opened 2021-11-17
Multiple documents using the same websocket connection


ChasLui:
The current document requires a websocket connection, and there is a limit to the number of browser connections

PR #484, opened 2023-01-19
feature/multiplexing

janthurau:
Not ready at all - WIP! refs #249, #283, #343, #455

Feedback & Contributions are more than welcome :-)

**The idea:**
* Extend the current messaging protocol and add a field at the beginning that indicates the documentName. We can later on even add another field to indicate the yjs subdoc Id.
* In order to keep most of the Hocuspocus server as it is, a websocket connection can now have multiple `Connection` instances that each only reacts to messages affecting them (by checking the documentName). `Connection`s are spawned by another onMessage listener on the WebSocket, so make sure that all messages are handled.
* Provider: not sure yet. Either the provider gets separated from the Websocket (in that case, two open documents would require two providers that each get the same WebSocket injected. Would also open the possibility of just injecting an existing Websocket connection), but that looked rather complicated. An alternative might be just allowing to pass multiple ydocs to the same provider.

**Protocol:**

Before:
`[HocuspocusMessageType][YjsMessageType][yjs binary]`

After:
`[documentName][HocuspocusMessageType][YjsMessageType][yjs binary]`

**Current Status:**
* Both provider and server are adjusted to include the current documentName. Syncing seems to work fine

**Current ToDos:**

- [x] Adjust the provider to allow opening a second document
- [x] The server needs to listen for a CLOSE message. This can be be sent by the provider to close a document without closing the entire connection
- [ ] The server should emit an 'ready' message once the connection is established, so that providers know it's safe to send first sync step
- [x] Check & add tests (e.g. for new CLOSE message)
- [x] Check all hooks - e.g. onStoreDocument, does the debounce work properly with different documents?
- [ ] Check all extensions (e.g. extension-redis - as long as hooks work fine it should be good)
- [x] Update broadcast messages
- [ ] ... ?

**Planned release:**
v2.0.0

** Breaking changes: **
* All yjs messages are now prefixed with the name of the document (including redis messages)
* All methods related to the connection have been moved to the providerwebsocket instance (e.g. disconnect(), connect(), etc)


raineorshine:
Thanks for opening this WIP!

My use case is a large graph with 1 million+ nodes represented as top-level documents with unique room names for flexible lazy loading.

> * Provider: not sure yet. Either the provider gets separated from the Websocket (in that case, two open documents would require two providers that each get the same WebSocket injected. Would also open the possibility of just injecting an existing Websocket connection), but that looked rather complicated. An alternative might be just allowing to pass multiple ydocs to the same provider.

Most of the YJS providers assume one document per provider: `y-webrtc`, `y-websocket`, and `y-indexeddb` among others. The user is expected to create a separate provider instance for each document, though it should be noted that this is more out of necessity than out of a deeply thought out design. [y-leveldb](https://github.com/yjs/y-leveldb#use-it) appears to be an exception, as all document operations are keyed by `docName`.

In hocuspocus, all of the provider configuration options are related to the Websocket connection. Only `document` (and `awareness`?) are document-specific. 

The provider events are much more split:

Websocket connection events:

- open
- connect
- close
- disconnect
- destroy

Doc-specific events:

- authenticated
- authenticationFailed
- message
- outgoingMessage
- synced
- awarenessUpdate
- awarenessChange

With multiple docs per provider, the config and events API's are largely preserved. You would simply add an additional `docname` argument to each of the doc-related events for subscribers. One document per provider would require larger changes to the API. Websocket connection config and events would need to be handled separately, but still easily shared among provider instances.

The downside of multiple docs per provider is that you then need to internally manage a list of docs, including add/remove, subscribe/unsubscribe, and connect/disconnect.

Personally I went in circles and didn't come to a conclusion, but hopefully sharing my thought process helps your thought process in some way :).

raineorshine:
I've started testing this branch in my app. Thanks again for all your work!

First thing I am noticing is that `HocuspocusProvider.on('status', ...)` is passing `{ event: { status } }` to my handler instead of `{ status }`. Is this expected?

`HocuspocusProviderWebsocket.on('status', ...)` passes `{ status }`.

janthurau:
@raineorshine thanks! That should not happen, seems I mixed something up as I remember explicitely adding this 🤔 Have you progressed further or shall I check/fix that first?

raineorshine:
I'm going to do some more testing. I'm getting an issue where only part of a transaction is completing and I need to isolate it.

raineorshine:
> @raineorshine thanks! That should not happen, seems I mixed something up as I remember explicitely adding this 🤔 Have you progressed further or shall I check/fix that first?

```patch
diff --git a/packages/provider/src/HocuspocusProvider.ts b/packages/provider/src/HocuspocusProvider.ts
index 29b58e40..068a7f67 100644
--- a/packages/provider/src/HocuspocusProvider.ts
+++ b/packages/provider/src/HocuspocusProvider.ts
@@ -149,25 +149,25 @@ export class HocuspocusProvider extends EventEmitter {
     this.on('authenticationFailed', this.configuration.onAuthenticationFailed)
 
     this.configuration.websocketProvider.on('connect', this.configuration.onConnect)
-    this.configuration.websocketProvider.on('connect', (e: Event) => this.emit('connect', { event: e }))
+    this.configuration.websocketProvider.on('connect', (e: Event) => this.emit('connect', e))
 
     this.configuration.websocketProvider.on('open', this.onOpen.bind(this))
-    this.configuration.websocketProvider.on('open', (e: Event) => this.emit('open', { event: e }))
+    this.configuration.websocketProvider.on('open', (e: Event) => this.emit('open', e))
 
     this.configuration.websocketProvider.on('message', this.onMessage.bind(this))
 
     this.configuration.websocketProvider.on('close', this.onClose.bind(this))
     this.configuration.websocketProvider.on('close', this.configuration.onClose)
-    this.configuration.websocketProvider.on('close', (e: Event) => this.emit('close', { event: e }))
+    this.configuration.websocketProvider.on('close', (e: Event) => this.emit('close', e))
 
     this.configuration.websocketProvider.on('status', this.onStatus.bind(this))
-    this.configuration.websocketProvider.on('status', (e: Event) => this.emit('status', { event: e }))
+    this.configuration.websocketProvider.on('status', (e: Event) => this.emit('status', e))
 
     this.configuration.websocketProvider.on('disconnect', this.configuration.onDisconnect)
-    this.configuration.websocketProvider.on('disconnect', (e: Event) => this.emit('disconnect', { event: e }))
+    this.configuration.websocketProvider.on('disconnect', (e: Event) => this.emit('disconnect', e))
 
     this.configuration.websocketProvider.on('destroy', this.configuration.onDestroy)
-    this.configuration.websocketProvider.on('destroy', (e: Event) => this.emit('destroy', { event: e }))
+    this.configuration.websocketProvider.on('destroy', (e: Event) => this.emit('destroy', e))
 
     this.awareness.on('update', () => {
       this.emit('awarenessUpdate', { states: awarenessStatesToArray(this.awareness.getStates()) })
```

raineorshine:
I encountered an issue where it doesn't sync when there is an `onAuthenticate` hook, even if it is empty.

```ts
const server = Server.configure({
  onAuthenticate: async () => {}
})
```

This occurs when there is one `HocuspocusProviderWebsocket` and multiple `HocuspocusProvider` instances. The websocket is authenticated, but updates from one client don't get synced to others. I checked out hocuspocus `v1.1.0` and it works as expected.

janthurau:
hey @ChasLui, @ahoyahoy, @DanSnow, @HelloAlexPan, @martinboksa, @lele0108, @shahriar-shojib,

mentioning all of you because you’ve asked for multiplexing support in the past. If you don’t need this anymore, sorry, you can stop reading :-)

I’ve just released a 2.0.0-alpha.0 pre-release to npm, which includes multiplexing (i.e. using one websocket connection for synchronization of multiple documents). The status is still a bit early, but I already wanted to gather as much feedback as possible and find potential testers in the community. I currently do not have a bigger project where I can test this, so if you can use multiplexing in your project and are willing to give the current implementation a try, this would be much appreciated and helps to get this merged.

On the frontend side, all connection logic has been split to a new ‘HocuspocusProviderWebsocket’, which needs to be passed to your HocuspocusProvider(s). So if you want to synchronize three documents, you’ll create one `HocuspocusProviderWebsocket` and three `HocuspocusProvider`, which gets passed the websocket (`webSocket` param). Authentication works per document, so the URL has to be passed to the HocuspocusProviderWebsocket, tokens to the provider related with the document. A working example can be found in `playground/frontend/src/pages/Index.vue`. Provider events are filtered by documentName, so the frontend shouldnt require further changes. Same on the backend, note that internally we’re spawning different `Connection` instances that each handle one document connection (through the same websocket). If you are somehow parsing the websocket messages yourself, please note that the protocol has changed and now includes the documentName at the first segment of the message.

You can find the PR here for details: https://github.com/ueberdosis/hocuspocus/pull/484

If you need any help, feel free to ping me on Discord or send me a direct message. We can even jump on a quick call if you are willing to help here. 

NPM: https://www.npmjs.com/package/@hocuspocus/server/v/2.0.0-alpha.0
Github Release: https://github.com/ueberdosis/hocuspocus/releases/tag/v2.0.0-alpha.0

husseinraoouf:
First Great work on the feature

I am testing this with an lexical and unfortunately it is not working because the lexical collaboration plugin calls `connect` on the yjs provider and the `HocuspocusProvider` doen't have `connect` now

I think the Provider should have the `connect` method that just call connect of the undelying HocuspocusWebsocketProvider if not connected yet or does noting if already connected

Thank you

PR #489, opened 2023-01-26
Update contributing.md testing instructions

raineorshine:
I have updated the test instructions. 

I'm not sure what `npm run test:open` was referring to, as the only test script in the package.json is `test`. Perhaps the testing infrastructure changed and the contributing.md was not updated.


mkriegeskorte (approved):
@raineorshine  Good catch! Thank you :)

PR #506, opened 2023-02-03
Fix bugs in webhook extension 

### LINKED ISSUE #505: Webhook fires ondisconnect even if not in events
**Description**
Webhook fires ondisconnect even if not in configured events

Offending code:
```
  async onDisconnect(data: onDisconnectPayload) {
    if (!this.configuration.events.includes(Events.onConnect)) {
      return
    }
```

fredpedersen:
Fixes #505

### DIFF
--- a/packages/extension-webhook/src/index.ts
+++ b/packages/extension-webhook/src/index.ts
@@ -182,16 +182,21 @@ export class Webhook implements Extension {
   }
 
   async onDisconnect(data: onDisconnectPayload) {
-    if (!this.configuration.events.includes(Events.onConnect)) {
+    if (!this.configuration.events.includes(Events.onDisconnect)) {
       return
     }
 
-    await this.sendRequest(Events.onDisconnect, {
-      documentName: data.documentName,
-      requestHeaders: data.requestHeaders,
-      requestParameters: Object.fromEntries(data.requestParameters.entries()),
-      context: data.context,
-    })
+    try {
+        await this.sendRequest(Events.onDisconnect, {
+          documentName: data.documentName,
+          requestHeaders: data.requestHeaders,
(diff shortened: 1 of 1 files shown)


packages/extension-webhook/src/index.ts
```diff
@@ -182,16 +182,21 @@ export class Webhook implements Extension {
...
-      context: data.context,
-    })
+    try {
+        await this.sendRequest(Events.onDisconnect, {
+          documentName: data.documentName,
+          requestHeaders: data.requestHeaders,
+          requestParameters: Object.fromEntries(data.requestParameters.entries()),
+          context: data.context,
+        })
+    } catch (e) {
```
janthurau, line 196: I think we should probably move this part to the Hocuspocus instance, so we don't have to take care of potential errors in the hooks

janthurau (changes requested):
thanks for your PR! 🙏

packages/extension-webhook/src/index.ts
carecodex: Good point, although it would be useful to differentiate between 403 errors for example and 500. I've not had time to delve into the working of the hocuspocus instance yet, but I wonder if it could be done in a way that fixes #508 too?

fredpedersen:
Delving into the code further, it appears that the hocuspocus server will allow all connections, only rejecting it once an onConnect event from the webhook is rejected. This leaves security vulnerabilities if for any reason the webhook response is delayed.

For this reason, in it's current state I don't think auth should be done using the webhook extension at all, instead you can use the normal onAuthenticate hook in server.

I have removed the code regarding auth errors, and only left the line fixing disabling onDisconnect.

janthurau (approved):

janthurau:
Thanks! 🥳

PR #511, opened 2023-02-06
chore: synchronize package-lock.json

Deckluhm:
I think `npm i` wasn't run after https://github.com/ueberdosis/hocuspocus/commit/cb21edca148bb63fcd890e594d789637f65712c0, causing the lockfile to become desynchronized/outdated.


janthurau (approved):
Thanks!

Issue #510, opened 2023-02-06
`@hocuspocus/transformer` and `@tiptap/core@2.0.0-beta.212` are incompatible


Deckluhm:
**Description**

`@hocuspocus/transformer` is referring to a path that no longer exists in `@tiptap/core@2.0.0-beta.212` (and it probably started with the `2.0.0-beta.210` refactoring).

To be exact, this is the import that is now failing because `@tiptap/core/dist/packages` no longer exists.

https://github.com/ueberdosis/hocuspocus/blob/main/packages/transformer/src/Tiptap.ts#L2

**Steps to reproduce the bug**

Instructions are in this repro README: https://github.com/Deckluhm/hocuspocus-tiptap-core-212

**Expected behavior**

It should compile without any error, even if using latest `@tiptap/core`.

**Environment?**
 - operating system: **Ubuntu 22.10**
 - hocuspocus version: `1.0.1`

PR #512, opened 2023-02-07
fix: make `transformer` compatible with `@tiptap/core@^2.0.0-beta.210`

Deckluhm:
Fixes https://github.com/ueberdosis/hocuspocus/issues/510

Ideally it would be great to merge https://github.com/ueberdosis/hocuspocus/pull/511 first.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4640,9 +4640,9 @@
       "license": "MIT"
     },
     "node_modules/@tiptap/core": {
-      "version": "2.0.0-beta.215",
-      "resolved": "https://registry.npmjs.org/@tiptap/core/-/core-2.0.0-beta.215.tgz",
-      "integrity": "sha512-1f0/ln/nhspKqykaoSSAK1ftm4XtKw7NyAjRQIUmEjAP0IxZKTBy/ShdwXl2/5xefrx+oaO6D5kVBjMZM0ioAw==",
+      "version": "2.0.0-beta.217",
+      "resolved": "https://registry.npmjs.org/@tiptap/core/-/core-2.0.0-beta.217.tgz",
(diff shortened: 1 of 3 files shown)


janthurau:
@Deckluhm thanks for this! There seem to be quite a few issues right now on TipTap-Side, so if possible I'd just wait a bit more and then directly move to the fixed version (probably once https://github.com/ueberdosis/tiptap/pull/3720 is merged)

Deckluhm:
I wasn't aware of this Tiptap issue since `@tiptap/core@^2.0.0-beta.210` is working fine on my side with this PR.

If the issue is impacting `@tiptap/core`, we can definitely wait for the PR to be merged :+1:

Deckluhm:
It got merged a few hours ago so I just updated my PR with `@tiptap/core@^2.0.0-beta.214` (which is the minimum required version to get the fix).

svenadlung:
@Deckluhm I think 214 should restore the old path 😂 But still I think your way of importing is even better. 
@bdbch what do you think?

bdbch:
I thought about this for a few minutes and I think having those types imported directly from the project would be cleaner and/or easier to use and discover.

Example:

```
import { Extensions } from '@tiptap/core
```

or

```
import { Extensions } from '@tiptap/core/types'
```

but I'd stick with the first one. I think this should already work with a lot of existing typings but it seems some are not exported from the entryfile / entry types file.

bdbch:
Obv. we would need to change this in the @tiptap/core package though.

Deckluhm:
> Obv. we would need to change this in the @tiptap/core package though.

`import { Extensions } from '@tiptap/core` is already possible with current `@tiptap/core`.

Deckluhm:
Is there anything preventing the merge of this PR?

janthurau (approved):
@Deckluhm Thanks! I've just pushed an update to .217 and will merge soon :)

Issue #279, opened 2022-01-07
Send stateless messages over the existing websocket connection


ViktorQvarfordt:
**The problem I am facing**

In collaborative applications one typically wants to send arbitrary stateless messages, for example to trigger an action on the server or on other connected clients. Hocupocus already establishes a websocket connection for synchronizing YDoc and YAwareness but both of these are stateful types of communication. Setting up a separate service for stateless messages would be unfortunate when we already have the websocket connection established by Hocuspocus.

**The solution I would like**

Ability to send client-to-server messages and broadcast messages. A broadcast message reaches all connected clients and the server, and can be initiated by either a client or the server.

For this we need a namespace but we already have a natural namespace identified by document name so I suggest we stick to that and keep it simple. (This may need to be coordinated with the thinking of subdocuments.)

Very rough usage sketch:

```ts
const provider = new HocuspocusProvider(..)
provider.broadcastMessage(data) // This message reaches all connected clients and server
provider.sendServerMessage(data) // This message reaches only the server and is initiated by a client
provider.on('message', (data, source) => ..) // source indicates a client id or server that initiated the message
// data should perhaps simply follow the websocket standard https://developer.mozilla.org/en-US/docs/Web/API/WebSocket/send#parameters.
// It will be common for users to send json data, but the user can JSON.stringify/parse manually
```

The naming needs to be figured out. Hocuspocus currently has another event with the same name `.on('message')`.

This feature should scale horizontally together with the upcoming Redis scaling mechanism.

**Alternatives I have considered**

Setting up a completely separate websocket server and sending stateless messages through this service.

**Additional context**

I think this feature fits well in the overarching vision of Hocuspocus as I understand it: To provide the backend solution for collaborative and real-time applications. I believe avoiding separate infrastructure for sending stateless messages would be a big win for many users of Hocuspocus.

janthurau:
Added in v1.1.0.

adrice727:
Has anyone had success sending messages from a `provider`?  `document.broadcastStateless()` works fine, but when I call `provider.sendStateless()`, I don't receive the message on the server or in other providers.  I'm on version `1.1.1` for all `@hocuspocus` packages.

janthurau:
@adrice727 can you share your code? I just tested, for me it works.

provider.sendStateless('test')

server:
``` 
async onStateless(data) {
    console.log(data)
  },
```

adrice727:
[...]

Issue #148, opened 2021-08-10
Add server instance method to close all open connections (support Blue/Green deployments)


jamesopti:
**The problem I am facing**
When a new code release of Hocuspocus goes out, I need to be able to close all open Hocuspocus connections to the old instance and force them to reconnect to the new one. This is necessary to ensure that clients can only connect to 1 Hocuspocus instance at a time for a given doc.

We're using AWS blue/green deployments to cutover traffic, but the websocket connections are not dropped, allowing existing clients to remain connected to the old instance and new clients to the new one.

**The solution I would like**
An instance method for Hocuspocus to close all open connections. `Server.closeAllConnections`

This is slightly different from `Server.destroy`, which would shutdown the http server and prevent a rollback from being possible.

**Alternatives I have considered**
We've worked around similar problems in the past using a reference to the instance and reaching into its internals:

```ts
this.getInstance().documents.forEach((document) => {
  document.connections.forEach((conn: Connection) => {
    conn.connection.close()
  })
})
```

**Additional context**
It's possible that this is solved later down the road once the Hocuspocus scaling work is complete (#87).

PR #516, opened 2023-02-14
feature: send stateless messages over the existing websocket connection

solirpa:
# Changes

-   Adds a new MessageType [Stateless]
	- The provider and server can send stateless messages to each other
	- The server can broadcast stateless messages to all connections based on document
-   Adds tests for the stateless message

# Context

Addresses [#148](https://github.com/ueberdosis/hocuspocus/issues/279)

### DIFF
--- a/docs/api/hooks/on-stateless.md
+++ b/docs/api/hooks/on-stateless.md
@@ -6,7 +6,7 @@ tableOfContents: true
 
 ## Introduction
 
-The `onStateless` hook is called after the server is received a stateless message. It should return a Promise.
+The `onStateless` hook is called after the server has received a stateless message. It should return a Promise.
 
 ## Hook payload
 
@@ -29,7 +29,7 @@ import { Server } from '@hocuspocus/server'
 const server = Server.configure({
   async onStateless(data) {
     // Output some information
-    console.log(`Server is received a stateless message "${data.payload}"!`)
+    console.log(`Server has received a stateless message "${data.payload}"!`)
   },
 })
 
--- a/docs/api/hooks/on-stateless.md
+++ b/docs/api/hooks/on-stateless.md
@@ -27,9 +27,13 @@ const data = {
 import { Server } from '@hocuspocus/server'
 
 const server = Server.configure({
-  async onStateless(data) {
+  async onStateless({ payload, document, connection }) {
     // Output some information
-    console.log(`Server has received a stateless message "${data.payload}"!`)
+    console.log(`Server has received a stateless message "${payload}"!`)
+    // Broadcast a stateless message to all connections based on document
+    document.sendStateless('This is a broadcast message.')
+    // Send a stateless message to a specific connection
+    connection.sendStateless('This is a specific message.')
   },
 })
 
--- a/docs/api/hooks/on-stateless.md
+++ b/docs/api/hooks/on-stateless.md
@@ -1,0 +1,37 @@
+---
+tableOfContents: true
+---
+
+# onStateless
+
+## Introduction
+
+The `onStateless` hook is called after the server is received a stateless message. It should return a Promise.
+
+## Hook payload
+
+The `data` passed to the `onListen` hook has the following attributes:
+
+```js
+const data = {
+  connection: Connection,
+  documentName: string,
+  document: Document,
+  payload: string,
+}
+```
+
+## Example
+
+```js
+import { Server } from '@hocuspocus/server'
+
+const server = Server.configure({
+  async onStateless(data) {
+    // Output some information
+    console.log(`Server is received a stateless message "${data.payload}"!`)
+  },
+})
+
+server.listen()
+```
--- a/docs/api/hooks/on-stateless.md
+++ b/docs/api/hooks/on-stateless.md
@@ -31,7 +31,7 @@ const server = Server.configure({
     // Output some information
     console.log(`Server has received a stateless message "${payload}"!`)
     // Broadcast a stateless message to all connections based on document
-    document.sendStateless('This is a broadcast message.')
+    document.broadcastStateless('This is a broadcast message.')
     // Send a stateless message to a specific connection
     connection.sendStateless('This is a specific message.')
   },
--- a/packages/server/src/Document.ts
+++ b/packages/server/src/Document.ts
@@ -215,6 +215,15 @@ export class Document extends Doc {
 
     return this
   }
+
+  /**
+   * Broadcast stateless message to all connections
+   */
+  public sendStateless(payload: string): void {
+    this.getConnections().forEach(connection => {
+      connection.sendStateless(payload)
+    })
+  }
 }
 
 export default Document
--- a/packages/server/src/Document.ts
+++ b/packages/server/src/Document.ts
@@ -219,7 +219,7 @@ export class Document extends Doc {
   /**
    * Broadcast stateless message to all connections
    */
-  public sendStateless(payload: string): void {
+  public broadcastStateless(payload: string): void {
     this.getConnections().forEach(connection => {
       connection.sendStateless(payload)
     })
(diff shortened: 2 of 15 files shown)


docs/api/hooks/on-stateless.md
```diff
@@ -0,0 +1,37 @@
+---
+tableOfContents: true
+---
+
+# onStateless
+
+## Introduction
+
+The `onStateless` hook is called after the server is received a stateless message. It should return a Promise.
```
janthurau, line 9: `has received`

docs/api/hooks/on-stateless.md
```diff
@@ -0,0 +1,37 @@
...
+
+## Example
+
+```js
+import { Server } from '@hocuspocus/server'
+
+const server = Server.configure({
+  async onStateless(data) {
+    // Output some information
+    console.log(`Server is received a stateless message "${data.payload}"!`)
```
janthurau, line 32: `has received`

janthurau:
@solirpa Thanks a lot for this!

It looks fine from my side (apart from the two typos), anything that you @jamesopti or @ViktorQvarfordt would like to add / see? Anything else that you need for your usecase? If not, LGTM :-)

janthurau (approved):

janthurau (dismissed):

packages/server/src/Document.ts
```diff
@@ -215,6 +215,15 @@ export class Document extends Doc {
 
     return this
   }
+
+  /**
+   * Broadcast stateless message to all connections
+   */
+  public sendStateless(payload: string): void {
```
jamesopti, line 222: It would be nice to have an optional argument here to send a message to a specific connection instead of all connections.

That would support the simple PING/PONG use case, as well as others where a single connection wants to message back and forth with the server.

I suppose if `connection.sendStateless` is a public method though, that this is already possible?

HengCC:
I have mentioned a similar issue before, I think this is a great PR,   in actual use. There are still many scenarios that need to broadcast other clients through the ws server. I think this broadcast method should be a public method on the server instance. At the same time, it should also be considered How to broadcast when deploying multiple servers. For example, compatibility with existing extension-redis.

packages/server/src/Document.ts
solirpa: If you want to send a message to a specific connection, you can search for the corresponding connection from the `document.connections` and then call `connection.sendStateless`.

And then, if you want to reply to a message sent by a certain provider, you can use the `connection` parameter in the onStateless hook.

I add some test cases.

solirpa:
> I have mentioned a similar issue before, I think this is a great PR, in actual use. There are still many scenarios that need to broadcast other clients through the ws server. I think this broadcast method should be a public method on the server instance. At the same time, it should also be considered How to broadcast when deploying multiple servers. For example, compatibility with existing extension-redis.

@HengCC Thank you for your suggestion.

The connections to the server are maintained in the `Document`, so I think the `broadcast` method is more intuitive in the `Document`.

The compatibility with extension-redis, I haven't come up with a more elegant solution.

You can publish a Redis broadcast message with the `payload` before calling `sendStateless`, and when other server receive the redis message, they can call the `sendStateless` method of their own Document.

HengCC:
> > I have mentioned a similar issue before, I think this is a great PR, in actual use. There are still many scenarios that need to broadcast other clients through the ws server. I think this broadcast method should be a public method on the server instance. At the same time, it should also be considered How to broadcast when deploying multiple servers. For example, compatibility with existing extension-redis.
> 
> @HengCC Thank you for your suggestion.
> 
> The connections to the server are maintained in the `Document`, so I think the `broadcast` method is more intuitive in the `Document`.
> 
> The compatibility with extension-redis, I haven't come up with a more elegant solution.
> 
> You can publish a Redis broadcast message with the `payload` before calling `sendStateless`, and when other server receive the redis message, they can call the `sendStateless` method of their own Document.

I think if you want to be compatible with extension-redis, a feasible solution is this. Before the server prepares to broadcast messages, you need to determine whether it is currently integrated through redis. If so, you may need to modify the source code of extension-redis. First, like Send an event in redis, subscribe to this event in extension-redis, and then broadcast it to all clients on the current server. If the redis extension is not integrated, just broadcast it directly.

solirpa:
> > > I have mentioned a similar issue before, I think this is a great PR, in actual use. There are still many scenarios that need to broadcast other clients through the ws server. I think this broadcast method should be a public method on the server instance. At the same time, it should also be considered How to broadcast when deploying multiple servers. For example, compatibility with existing extension-redis.
> > 
> > 
> > @HengCC Thank you for your suggestion.
> > The connections to the server are maintained in the `Document`, so I think the `broadcast` method is more intuitive in the `Document`.
> > The compatibility with extension-redis, I haven't come up with a more elegant solution.
> > You can publish a Redis broadcast message with the `payload` before calling `sendStateless`, and when other server receive the redis message, they can call the `sendStateless` method of their own Document.
> 
> I think if you want to be compatible with extension-redis, a feasible solution is this. Before the server prepares to broadcast messages, you need to determine whether it is currently integrated through redis. If so, you may need to modify the source code of extension-redis. First, like Send an event in redis, subscribe to this event in extension-redis, and then broadcast it to all clients on the current server. If the redis extension is not integrated, just broadcast it directly.

The server should not be responsible for checking whether the redis-extension is integrated, as this would be invasive for the server.

Instead, the server should provide a method that notifies the extension when it is broadcasting a message, for example, a `beforeBroadcastStateless` hook.

However, I am not currently able to handle the case of integrating with the redis-extension. Perhaps in the future, if there is a need, I will submit another PR to address this.

solirpa:
@janthurau @jamesopti 
I changed the API name of the broadcast message. 

Please review it again.

janthurau:
thx @solirpa! I'm just wondering what do to with the Redis extension, I'll play around this week and check if I can easily make it forward these messages as well (I guess that shouldn't be too hard), and then we can get this merged.

As we're officially supporting Redis, it would feel weird to not support this.

janthurau (approved):

janthurau:
thanks @solirpa! Merging this. New PR from me is coming, I've added redis support.

solirpa:
Thank you @janthurau 

I created a new [branch](https://github.com/solirpa/hocuspocus/pull/2/files)  that is compatible with the redis-extension, but I'm not sure if there is anything missing.

I think it could be a reference for your new pull request.

janthurau:
thx! We both worked on it now, I've created https://github.com/ueberdosis/hocuspocus/pull/522 :-) Will merge them together soon, on short look it seems we did exactly the same haha

raineorshine:
> * The provider and server can send stateless messages to each other

Great feature! Can you give an example of how the provider can send a stateless message to the server? I only see documentation for server → client.

janthurau:
@raineorshine `provider.sendStateless('any string')`

indeed that's missing in docs, will note!

raineorshine:
Got it, thanks.

However, why send a string instead of JSON? There is a multi-decade precedent for JSON as an interchange format for stateless protocols. I think hocuspocus should handle encoding of stateless messages just like it handles encoding of yjs messages.

janthurau:
this is sent using the same protocol as all other messages, so encoded as binary. 

Yjs sends awareness updates using writeVarString(JSON.stringify(state)), which is the same thing you can do here: `provider.sendStateless(JSON.stringify(customObject))`

Let me know if that makes sense! Haven't released this feature yet, as compatibility for extension-redis is still pending (and probably some docs updates).

raineorshine:
Sure, that makes sense. I was just looking for a way to avoid making the user encode/decode JSON, but maybe that's not possible.

> this is sent using the same protocol as all other messages, so encoded as binary.

Yet, but by accepting only a string, you prevent the user from encoding actual binary data. It's just a binary representation of a string. (Maybe not all that important for the stateless protocol, just pointing out the limitation.)

PR #517, opened 2023-02-14
feature: support redis cluster for extension-redis

lzj723:
To deploy hocuspocus in a HA setup, it's highly recomended to support Redis Cluster, not only singleton redis.

Now it support four ways to create a pub/sub redis instance:
- host + port
- nodes for cluster
- duplicate from an existed redis instance
- create from `createClient ` method

### DIFF
--- a/packages/extension-redis/src/Redis.ts
+++ b/packages/extension-redis/src/Redis.ts
@@ -19,7 +19,7 @@ import {
 } from '@hocuspocus/server'
 import kleur from 'kleur'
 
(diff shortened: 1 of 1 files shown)


janthurau (commented):
Thanks @lzj723! 🙏

janthurau:
@lzj723 Do you happen to have any comments about https://github.com/ueberdosis/hocuspocus/pull/453 (updating ioredis to the latest version, but dropping / replacing node-redlock as they seem unmaintained)

janthurau (approved):

Issue #509, opened 2023-02-06
Content duplication when other users connect.


binurajrvt:
**Description**
I am building a collaborative editor with TipTap and Hocuspocus with React as shown in the examples in the documentation. After initialisation, I check the number of connected users, and if it is only one (editor.storage.collaborationCursor.users.length), I set the content fetched from a Laravel API (means its an existing document saved in db). Everything works fine till this point. But when a second user logs in, the document content is duplicated across all editor instances. For testing, I am using sqlite driver in hocuspocus server. When other users also connect the content is duplicated just as multiplications of the number of users.

**Steps to reproduce the bug**
Steps to reproduce the behaviour: 
1. After initialisation, set the content
2. Login as another user in another browser or incognito.
3. The content gets duplicated on all instances

**Expected behaviour**
Need to set the initial content from a API, if no users have currently opened the document, and it does not get duplicated if other users logs in.

**Environment?**
 - operating system: MAC OS 12.1, Ubuntu
 - browser: Chrome, Firefox, Safari
 - mobile/desktop: Desktop
 - hocuspocus version: 1.0.0-beta.2

fredpedersen:
Could you explain why you're getting the initial content from an API? That's not really how hocuspocus is designed to work, you can set the inital document if it's never been opened before but after that it should really be persisted in binary form and retreived onloaddocument.

This thread might be of use #344

binurajrvt:
Thank you for the response. I am saving the content as html in MySQL db and from there we convert it to XML and supply  for another projects as response to an API request from that project. We need the HTML format. 
"set the inital document if it's never been opened before" - is there any method to check if the document is not opened?
That means the initial content in binary format is to be in Hocuspocus SQLite/RocksDB only or should I save the content in binary format (with `Y.encodeStateAsUpdate(ydoc)`)  also in the API MYSQL  database where I save html format content also and fetch that data and set the initial  content with `Y.applyUpdate(ydoc, props.savedYDoc)` as seen somewhere. Kindly advice. Any code samples will be of great help.

binurajrvt:
Thank you again for the quick response. I will try with that and will post here. Again:

> "set the inital document if it's never been opened before"

How to make sure if its never opened?

fredpedersen:
If it's been opened, it will exist in your table as yjs doc via the database extension

janthurau:
[...]

Issue #344, opened 2022-05-26
HocusPocus appends content to editor when restarted


matsko:
**Description**
When HocusPocus connects from the front-end over to the backend via a websocket, and the backend is restarted, then HocusPocus will reconnect the socket, but will accidentally append the contents loaded by `onLoadDocument` hook into the TipTap editor.

The expected behavior here is that the contents would just sync (if they are the same).

**Steps to reproduce the bug**
1. Run `git clone https://github.com/matsko/hp-test-app`
1. Run `yarn install` in the `api` dir
1. Run `yarn install` in the `web` dir
1. Run `./server.sh` in one tab
1. Run `./web.sh` in one tab
1. Go to `localhost:3000`
1. See the page load and then close the `./server.sh` command
1. Rerun `./server.sh`
1. Now the text has been appended into the document

**Expected behavior**
The content that the server returns is the current state of the document and it should just sync up with what's in the user's browser.

**Screenshot, video, or GIF**
https://user-images.githubusercontent.com/93018/170562480-99ad059e-0ed5-4250-94ea-8d9d4e8e5d76.mov

**Environment?**
 - operating system: Mac osX Monterey
 - browser: Chrome
 - mobile/desktop: Desktop
 - hocuspocus version:  `1.0.0-alpha.102` (server) and `1.0.0-alpha.36` (provider)

**Additional context**
Note that there is another, unrelated, error showing up in the reproduction. This error didn't show up on my actual project and the actual error here is unrelated.

PR #518, opened 2023-02-15
Additional documentation on Laravel integration

fredpedersen:
Provide Laravel implementation docs for when the hocuspocus-laravel package isn't suitable. Also helps to avoid issues such as #509 and #344 by further explaining the need for separate primary storage in YJS binaries in addition to any format that Laravel can read.


janthurau (approved):
Thanks a lot @fredpedersen

PR #521, opened 2023-02-22
Add missing update field to onStoreDocumentPayload

daniel-johnsson:
Noticed that `update` was missing from the `onStoreDocumentPayload` interface. 

The documentation regarding the payload states that the update should be a part of the payload:
https://github.com/ueberdosis/hocuspocus/blob/main/docs/api/hooks/on-store-document.md

We can also see that the update is included in the payload from `handleDocumentUpdate`
https://github.com/ueberdosis/hocuspocus/blob/main/packages/server/src/Hocuspocus.ts#L530-L555

### DIFF
--- a/packages/server/src/types.ts
+++ b/packages/server/src/types.ts
@@ -220,6 +220,7 @@ export interface onStoreDocumentPayload {
   instance: Hocuspocus,
   requestHeaders: IncomingHttpHeaders,
   requestParameters: URLSearchParams,
+  update: Uint8Array,
   socketId: string,
 }
 


janthurau:
hey @daniel-johnsson, thanks for this!

This is a bit dangerous, as the `onStoreDocument` hook actually can be triggered once for multiple updates because it's denounced, so I actually was already thinking to remove it completely from the payload.

Need to think about this..

daniel-johnsson:
Hi @janthurau , I thought this was the solution to my problem, however as you say the update will trick you. I think it should be removed completely from the `onStoreDocument`, would be cool though if it would be possible to pass to merge all debounced updates or pass all in an array. This looks tricky to achieve tough. 

Some context why I wanted the update:
I noticed that `onStoreDocument` where triggered if all connections to a document were closed. I have not found the root of this, but I figure (correct me if I am wrong) it's some safe keep to store the debounced changes. This would also trigger even if no changes to the document had been made. To prevent unnecessary stores I started to check if the update was defined. This ended up not working since per the issue you point out, this `onStoreDocument` call when all disconnect would always have an undefined update. Ended up scarping the idea of using the `update` and just added the last change time to document.meta that I compare to the db. This works but feels like it could be some race condition issues in the future.

Issue #559, opened 2023-03-30
Empty `requestParameters` in Webhook Extension Payload, after upgrading tiptap and hocuspocus from "2.0.0-beta.220" to "2.0.0"


fdietz:
**Description**
We've tried upgrading tiptap and hocuspocus from "2.0.0-beta.220" to "2.0.0". 

The Webhook Extension is not sending the "requestParameters" anymore. It is an empty object.

Are there any configuration changes, we need to apply in the new version? 

**Steps to reproduce the bug**
Upgrade and check the payload of the Webhook Extension.

**Expected behavior**
We expect the the "requestParameters" to include the Parameters passed along in the HocuspocusProvider as "parameters".

**Additional context**
We are a sponsor (fobizz | 101skills GmbH)

fdietz:
Hi @janthurau! And thanks you for looking into this.

I've noticed something else, but am unsure if it's a new feature. There's now a `token` param which can be passed by the HocuspocusProvider to the Server. I've tested this with the webhook, but it is also not passed along.

And the HocuspocusProvider does a new console.log `The provided authentication token isn’t allowed to connect to this server. Will try again`. Which is I assume because our webhook returns an unauthorized status (due to missing params).

Please, let me know if you need more information or if you need me to test something else.

Many thanks in advance

fdietz:
I've updated from "2.0.0-beta.220". I just never found the token in the documentation :-). Instead I passed our own data using the requestParameters.

The Webhook always passed the correct data of the individual user. We didn't use the hook you are referring to.

fdietz:
Just wanted to let you know that the `token` is not passed along by the webhook extension in version "2.0.0-beta.220" either. Assuming that when we pass it to the hocuspocus provider on the client, it would be passed through by the webhook extension.

Not to further confuse things with the `token` topic here, as the original problem was that the `requestParameters` are not passed through. And I'm unsure if it is related or something completely different.

janthurau:
@fdietz sorry for the delay, I'm expecting this to be fixed today.

janthurau:
@fdietz v2.0.2 is currently publishing, let me know if that fixes your issue :)

fdietz:
Hi @janthurau! That was quick!

Everything works again. The update went smoothly 🎉 

Thank you so much 🙏

PR #571, opened 2023-04-03
server can't get parameters.

lin52025i:
If the `HocuspocusProvider` does not provide a `websocketProvider`, the `parameters` of the `HocuspocusProvider` are not available to the server.


konstantin24121:
this one should fix #559

janthurau (approved):

janthurau:
Thanks @lin52025i!

PR #573, opened 2023-04-05
Delaying redis extension disconnect to allow last minute syncs to happen

### LINKED ISSUE #557: A quick disconnect after a change doesn't allow the change to be broadcasted through redis
**Description**
If a connection is closed quickly right after a change, the redis extension will not have enough time to broadcast the change properly. In fact the first publish sync message is sent, but because the redis subscription is removed in the onDisconnect hook of the redis extension, the further sync events are not published. Therefore the change never reaches the other clients through the redis channel.

1. Have two instance of the server running and sync them through the redis extension.
2. Make a change on a document in the first server from a browser and quickly close the connection (this has to be programmatically).
3. The corresponding document in the second server never receives the change.

**Expected behavior**
The document in the second server should receive the change.

ebads67:
fix #557 

Description
If a provider connection is closed quickly right after a change, the redis extension will not have enough time to broadcast the change properly. In fact the first publish sync message is sent, but because the redis subscription is removed in the onDisconnect hook of the redis extension, the further sync events are not published. Therefore the change never reaches the other clients through the redis channel.

Have two instance of the server running and sync them through the redis extension.
Make a change on a document in the first server from a browser and quickly close the connection (this has to be programmatically).
The corresponding document in the second server never receives the change.

Expected behavior:
The document in the second server should receive the change.

Note that this behavior is reproduced in the current test setup, as it uses a local redis instance. To reproduce this you need to connect to a remote redis instance.

### DIFF
--- a/packages/extension-redis/src/Redis.ts
+++ b/packages/extension-redis/src/Redis.ts
@@ -62,6 +62,11 @@ export interface Configuration {
    * The maximum time for the Redis lock in ms (in case it can’t be released).
    */
   lockTimeout: number,
+  /**
+   * A delay before onDisconnect is executed. This allows last minute updates'
+   * sync messages to be received by the subscription before it's closed.
+   */
+  disconnectDelay: number,
 }
 
 export class Redis implements Extension {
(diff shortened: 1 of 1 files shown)


janthurau:
😎  Thanks!!

janthurau (approved):

PR #595, opened 2023-04-30
Fixes Spelling Mistake

jjgmckenzie:
I'm not sure if this is an intentional mistake, to demonstrate that the development team is German? 

If not, then, here you go

Thank you for the amazing work on this and your other products.


janthurau:
Thansk! :-D

Issue #606, opened 2023-05-16
`connection` can be either Connection or ConnectionConfiguration


canadaduane:
**The problem I am facing**
It's a bit confusing that there is a `connection` key in various hook payloads that is either a `Connection` instance or a `ConnectionConfiguration` object. Both are quite different:

Connection
```ts
  webSocket: WebSocket
  context: any
  document: Document
  pingInterval: NodeJS.Timeout
  pongReceived = true
  request: HTTPIncomingMessage
  timeout: number
  callbacks: any
  socketId: string
  lock: AsyncLock
  readOnly: Boolean
  logger: Debugger
```

ConnectionConfiguration
```ts
  readOnly: boolean
  requiresAuthentication: boolean
  isAuthenticated: boolean
```

But they appear to be the same at first glance:

```ts
onConnect({connection}) {
  // connection is ConnectionConfiguration
})

beforeHandleMessage({connection}) {
  // connection is Connection
})
```

**The solution I would like**
It would be great if `ConnectionConfiguration` were always `connectionConfig` and `Connection` were always `connection`.

canadaduane:
I should add some context. I need to access the Connection object from the `connected` hook. I'm thinking about adding a PR to include it in the `connectedPayload`; however, `connected` already includes ConnectionConfiguration at a key called `connection`.

(At an even higher level of context: I am trying to add a custom "awareness" implementation, and need access to the websocket directly).

Issue #583, opened 2023-04-20
Subdocs


ImSingee:
Is there any plan to support yjs’s sub docs feature? It’a useful to develop a notes app (like the official example, use folder as root document).

https://docs.yjs.dev/api/subdocuments

janthurau:
hey @ImSingee, supports for subdocuments is planned, but probably towards the end of the year. We haven't started on this yet, so feel free to help if you can :-)

janthurau:
👍  let me know if you wanna discuss on Discord, I have some idea already (not sure if it would work though). There is also an idea here (not for hocuspocus, but it kinda shows the idea): https://discuss.yjs.dev/t/multiple-room-sync-subdocument/403/4

ArihantBapna:
Hi any updates on this?

dorukgunes:
Hi @janthurau are there any updates on this 🙋

MaxTwentythree:
I'm also interested, what's the state? I currently have to deciding on whether or not to use this package.. for my application I have to work with subdocuments (or a similar connect, because of granular permission) and as far as I understand it it's not supported right now, or is it?

PR #607, opened 2023-05-16
Add connection and context to `connected` hook payload

canadaduane:
Adds `connection` and `context` to the `connectedPayload` type for the `connected` hook.

Because `connectedPayload` already has `connection` (`ConnectionConfiguration`), if this PR intention is acceptable, we have the choice of either:

(a) make the most sensible change, but require a breaking change (i.e. bump the major version)
(b) make the least disruptive change, but add a key that could be a little confusing / misdirecting

This PR currently applies choice (a), but it could easily be switched. See #606.

### DIFF
--- a/packages/server/src/Hocuspocus.ts
+++ b/packages/server/src/Hocuspocus.ts
@@ -436,7 +436,12 @@ export class Hocuspocus {
         incoming.emit('message', input)
       })
 
-      this.hooks('connected', { ...hookPayload, documentName })
+      this.hooks('connected', {
+        ...hookPayload,
+        documentName,
+        context,
+        connectionInstance: instance,
+      })
     }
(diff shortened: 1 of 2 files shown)


janthurau:
Thanks for this!

Let's take the 2nd choice for now (no BC break), maybe we can just name it `connectionInstance`. The naming won't be that nice but I don't want to bump the major just because we rename one variable :) Maybe you can add an @TODO so we rename it in the next major (that might be with subdocs, #583). Happy to merge then!

canadaduane:
> maybe we can just name it `connectionInstance`

> add an @todo so we rename it in the next major

Done! Thanks for the feedback.

janthurau (approved):

PR #608, opened 2023-05-18
Feature/537 node resolution node16

janthurau:


reknih:
It would be great if this could be merged soon – we are currently porting our patch to each version of Hocuspocus to upgrade, this would be a huge relief!

janthurau:
@reknih I've just merged this and released - pls let me know if it works as intended :) I was a bit hesitant to do it as I'm not sure and this is hard to test without a release, but I just tested it out and it seems fine 🥳

reknih:
Thank you! I got around to updating the package now. In the current release, two imports fail because the old convention stuck around. I have created a PR (#647).

Issue #222, opened 2021-10-13
Ability to open a connection locally


georeith:
**The problem I am facing**

We have some additional services that want to make changes to the document but don't want to subscribe to changes on the document or even load the document.

For example we have a service that exports some of our documents and wants to set a boolean flag on the document that it is doing so. To make changes to the document currently you need to set up a websocket provider, connect to the server via websockets, load the document, make your change and then close the channel.

When this change is a simple flipping of a boolean value it would be neater if we could do it over HTTPS.

**The solution I would like**

This is almost possible by just using `express` as a front to `hocuspocus` and accessing `server.documents` from your `express` endpoint and making your change. The issue is this only works for documents currently open on the server.

I think it would be great to add a way to manually open a connection to a document locally, forcing hocuspocus to load the document into memory so that we can make changes to it from outside of hocuspocus.

I imagine the syntax could look something like this:

```ts
const connection = await server.createConnection(documentName); // open a connection to the document locally
const document = server.documents[documentName]; // grab the document that is now open
// do something with document
connection.close(); // release the local connection to the document
```

**Alternatives I have considered**

I have considered just having my endpoint retrieve the document from storage if it doesn't exist on the server, but it creates race conditions (what if hocuspocus opens the document sometime during this operation) and reduces flexibility (endpoint has to be aware of all the different storage and their order).

canadaduane:
I have a half-baked solution that I wanted to share, in the hopes we might make it a 'fully baked' solution:

```ts
const bindTransact =
  ({ context, instance }: { context?: any; instance: Hocuspocus }) =>
  async (documentName: string, transaction: (document: Y.Doc) => void) => {
    // We need to access a private function here, because documents are normally created by a
    // connection to the server. We *are* the server, however, so we don't need a connection.
    // eslint-disable-next-line dot-notation
    const document: Document = await instance["createDocument"](
      documentName,
      {},
      "server",
      {
        isAuthenticated: true,
        readOnly: false,
      },
      context,
    )
    transaction(document)

    const hookPayload = {
      clientsCount: document.getConnectionsCount(),
      context,
      document,
      documentName,
      instance,
      requestHeaders: {},
      requestParameters: {},
      socketId: "server",
    }

    // eslint-disable-next-line @typescript-eslint/no-floating-promises
    instance
      .hooks("onStoreDocument", hookPayload)
      .catch((error) => {
        if (error?.message) {
          throw error
        }
      })
      .then(() => {
[...]

PR #610, opened 2023-05-20
Add direct connections to documents

canadaduane:
This allows server code to create an in-memory direct "connection" to a document, without a websocket.

- Introduces `openDirectConnection` function on the Hocuspocus server
- There is now a `DirectConnection` class returned by this function, and `disconnect` can be called on it
- The main method to mutate a Y.Doc is via `directConnection.transact(document => { ... })`
- Adds tests

Addresses #222

### DIFF
--- a/packages/server/src/DirectConnection.ts
+++ b/packages/server/src/DirectConnection.ts
@@ -1,0 +1,53 @@
+import { URLSearchParams } from 'url'
+import Document from './Document'
+import type { Hocuspocus } from './Hocuspocus'
+
+export class DirectConnection {
+  documentPromise: Promise<Document> | null = null
+
+  instance!: Hocuspocus
+
+  context: any
+
+  document?: Document
+
+  /**
+   * Constructor.
+   */
+  constructor(
+    documentPromise: Promise<Document>,
+    instance: Hocuspocus,
+    context?: any,
+  ) {
+    this.documentPromise = documentPromise
+    this.instance = instance
+    this.context = context
+  }
+
+  async transact(transaction: (document: Document) => void) {
+    if (!this.documentPromise) {
+      throw new Error('direct connection closed')
+    }
+
+    this.document = await this.documentPromise
+    this.document.addDirectConnection()
+
+    transaction(this.document)
+
+    this.instance.storeDocumentHooks(this.document, {
+      clientsCount: this.document.getConnectionsCount(),
+      context: this.context,
+      document: this.document,
+      documentName: this.document.name,
+      instance: this.instance,
+      requestHeaders: {},
+      requestParameters: new URLSearchParams(),
+      socketId: 'server',
+    })
+  }
+
+  close() {
+    this.document?.removeDirectConnection()
+    this.documentPromise = null
+  }
+}
--- a/packages/server/src/DirectConnection.ts
+++ b/packages/server/src/DirectConnection.ts
@@ -3,34 +3,31 @@ import Document from './Document'
 import type { Hocuspocus } from './Hocuspocus'
 
 export class DirectConnection {
-  documentPromise: Promise<Document> | null = null
+  document: Document | null = null
 
   instance!: Hocuspocus
 
   context: any
 
-  document?: Document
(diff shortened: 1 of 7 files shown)


jjgmckenzie:
Looking forward to this! :+1:

packages/server/src/Document.ts
```diff
@@ -121,11 +124,25 @@ export class Document extends Doc {
     return this
   }
 
+  addDirectConnection(): Document {
```
janthurau, line 127: do we need this (`directConnectionsCount`) actually? 🤔

packages/server/src/DirectConnection.ts
```diff
@@ -0,0 +1,53 @@
...
+    documentPromise: Promise<Document>,
+    instance: Hocuspocus,
+    context?: any,
+  ) {
+    this.documentPromise = documentPromise
+    this.instance = instance
+    this.context = context
+  }
+
+  async transact(transaction: (document: Document) => void) {
```
janthurau, line 27: not sure I get the idea of this. How would this be used?

```typescript

const server = new Hocuspocus();

const myDocument = await server.getDirectConnection({documentName: 'my-document', context: {}})

await myDocument.transact((doc) => {
  doc.getMap('test').set('a', 'b');
});

// 2nd call would then again trigger addDirectConnection and await the documentPromise?
await myDocument.transact((doc) => {
  doc.getMap('test').set('a', 'b');
});
```

If I didn't get it wrong, I'd probably just make the `await server.getDirectConnection` set up everything (i.e. increase connection count, await document promise). `transact` method would then probably only call `storeDocumentHooks` after the transaction

packages/server/src/Hocuspocus.ts
```diff
@@ -903,6 +909,24 @@ export class Hocuspocus {
   getMessageLogs() {
     return this.debugger.get()?.logs
   }
+
+  getDirectConnection({ documentName, context }: { documentName: string, context?: any }): DirectConnection {
```
janthurau: maybe the signature can just be

`getDirectConnection(documentName: string, context?: any): DirectConnection`

(or probably `openDirectConnection`? 🤔)

janthurau (commented):
Thanks a lot for this! I've added a few thoughts, let me know what you think!

packages/server/src/DirectConnection.ts
janthurau: ah, I see you just pushed a fix while I was looking at it, which is doing exactly that :=) resolving!

packages/server/src/Document.ts
canadaduane: The scenario I'm guarding against here (I think) is:
- a direct connection is opened
- a regular connection is opened, modifications made, then closed
- the usual "clean this document out of memory" code (https://github.com/ueberdosis/hocuspocus/blob/main/packages/server/src/Hocuspocus.ts#L785) should not be triggered, because the direct connection still has a claim on the document

Testing this at https://github.com/ueberdosis/hocuspocus/pull/610/files#diff-f23cea543b7ce0921f130f63eae2ec74a5941c3486239317a010ee8a6b5970b8R4

packages/server/src/Hocuspocus.ts
canadaduane: Changed per your recommendation

janthurau (approved):

janthurau:
@canadaduane 😎  I've just pushed two more commits, one test case to make sure `onChange` is triggered properly (wasn't sure about that, but it works!), and I've replaced the last implementation of onStoreDocument with your new `onStoreDocumentHooks` method.

I think we're ready to merge!?

canadaduane:
Looks good to me! Thanks @janthurau

jjgmckenzie:
Hey @canadaduane - any chance you could give a quick example of how you have used this binding to edit documents? 

Do you perform edits to the YJS directly, or do you convert to another format and apply the edits back to it?

No worries if you're busy, I'll be able to figure it out :)

PR #613, opened 2023-05-22
failing test for readonly multiplexing

YousefED:
This unit test fails, indicating that when we connect with Multiplexing to 2 documents, both are registered as "readOnly", even though this property is only set on 1 of the documents.

I think this is because in HandleConnection ([link](https://github.com/YousefED/hocuspocus/blob/729b6c1ddfb5f12daceb242843e09d9e402c83ca/packages/server/src/Hocuspocus.ts#L374)), `connection` and `hookPayload` are set outside of the handlers, meaning that the `connection` object will always be the same. To be honest the logic here (with all the nested handlers and connections) is a bit difficult to follow, so I'm hesitant to look further into a fix


YousefED:
Closed by #631

PR #614, opened 2023-05-22
add tests for hasUnsyncedChanges

YousefED:
This PR adds unit tests for `hasUnsyncedChanges`.

It also surfaces an issue where the current implementation of `hasUnsyncedChanges` is broken:

The final test fails, because the implementation of `unsyncedChanges` depends on an update being "bounced back" after applied from the server to the client (source: https://github.com/YousefED/hocuspocus/blob/711f5de609a6de628700fc1ff85bfae795e24984/packages/provider/src/MessageReceiver.ts#L90). When receiving a `messageYjsUpdate`, it assumes this is the update the client sent to the server. However, the test demonstrates that the client could also be in read-only mode, and that the update can come from another client instead

### DIFF
--- a/tests/provider/hasUnsyncedChanges.ts
+++ b/tests/provider/hasUnsyncedChanges.ts
@@ -1,5 +1,6 @@
 import test from 'ava'
 import { retryableAssertion } from 'tests/utils/retryableAssertion'
+import * as Y from 'yjs'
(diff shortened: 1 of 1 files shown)


YousefED:
I've extended this PR with a test to also see whether the document has "initial" changes that are not synced. The current implementation also fails for this scenario

janthurau:
Closing this as #635 is the follow-up.

Issue #579, opened 2023-04-13
Multiplexing closes connection on auth failed


YousefED:
**Description**
I'm using the same WS connection to connect to multiple documents.

For document A, I don't have access (I provide an invalid token)
For document B, I do have access.

As soon as I connect to document A, the websocket connection is closed.

**Expected behavior**
In a multiplexing environment where auth is document based, I'd expect the architecture to be completely independent. Anything I do with document A should be unrelated to what I do with document B and vice versa

jinyang1994:
Is there any progress on the issue? @janthurau

janthurau:
not yet - this was expected behavior for the multiplexing release. It definitely makes sense to change this though - we're always open for contributions :)

PR #615, opened 2023-05-22
don't close when auth fails

YousefED:
This fixes and tests https://github.com/ueberdosis/hocuspocus/issues/579

There was a test for this already in place, but the issue only occured when first connecting with an invalid token, and then connecting with a valid token. In that case the connection wouldn't be established.

To resolve this, I decided to remove the closing behavior all together. For this, the other test ("stops when the onAuthenticate hook throws an Error") has also been updated

### DIFF
--- a/packages/server/src/ClientConnection.ts
+++ b/packages/server/src/ClientConnection.ts
@@ -273,19 +273,7 @@ export class ClientConnection {
           category: message.category,
         })
 
-        // Ensure that the permission denied message is sent before the
-        // connection is closed
-        this.websocket.send(message.toUint8Array(), () => {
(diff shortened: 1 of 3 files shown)


janthurau (approved):
👍  same here, just linting :)

janthurau:
I've just rebased this to the latest changes (ClientConnection etc.). Just not sure what I want to do with it, as it changes the behavior of errors in onLoadDocument (terminate connection vs. just sending 'permission denied'). Not sure if that might break something for someone? 🤔

janthurau:
ok, LGTM.

janthurau (approved):

PR #616, opened 2023-05-22
passes authorized scope to client

YousefED:
This passes the information on whether a connection is readonly or not to the client. The scope is implemented as a string sent on the "authenticated" message, so it's easily extensible in case we need to add different scopes later


janthurau (approved):
🙏 once the linting issue is fixed this is ready to merge :)

janthurau (approved):

PR #624, opened 2023-06-13
refactor: use async/await in HocusPocus.ts to make code more readable

YousefED:
This PR refactors Hocuspocus.ts to get rid of the `.then` and `.catch` constructs. Imo, this makes the code a lot more readable and will make it easier to do other improvements / refactors


janthurau (approved):

janthurau:
Thanks! 🥳

PR #626, opened 2023-06-14
Refactor: extract ClientConnection

YousefED:
This PR builds on top of #624

In this PR, we introduce a ClientConnection that is responsible for handling the connection with a single websocket client.

It extracts the logic from `handleConnection` and `createConnection` to a separate class.

After this, it will become a lot easier to follow the code flow and make refactors like cleaning up all the nesting that we now have in `ClientConnection.ts`. For example, we could consider extracting nested methods like `handleQueueingMessage`, `setupNewConnection` and `messageHandler` to become class methods. Or we can fix bugs like #613 

There are a few TODOs left to make the refactor nicer, mainly to raise events from `ClientConnection.ts` instead of calling passing `HocusPocus.hooks` directly. However, I wanted to make the PR as small as possible (making it easier to review), and imo this could already be merged

### DIFF
--- a/packages/server/src/ClientConnection.ts
+++ b/packages/server/src/ClientConnection.ts
@@ -1,0 +1,352 @@
+import { IncomingMessage } from 'http'
+import {
+  Forbidden, Unauthorized, WsReadyStates,
+} from '@hocuspocus/common'
+import * as decoding from 'lib0/decoding'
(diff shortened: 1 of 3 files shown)


janthurau (approved):

janthurau:
Thank you for your effort to make our code base better!! :-)

PR #631, opened 2023-06-16
Refactor/reduce nesting and fix hookPayload per document

### LINKED ISSUE #579: Multiplexing closes connection on auth failed

YousefED:
This refactor:
- reduces the nesting in ClientConnection.ts by extracting inline functions to class members
- fixes #579 (and adds the tests from https://github.com/ueberdosis/hocuspocus/pull/613), by making hookPayload document specific

### DIFF
--- a/packages/server/src/Hocuspocus.ts
+++ b/packages/server/src/Hocuspocus.ts
@@ -373,11 +373,11 @@ export class Hocuspocus {
    * … and if nothings fails it’ll fully establish the connection and
    * load the Document then.
    */
-  handleConnection(incoming: WebSocket, request: IncomingMessage): void {
+  handleConnection(incoming: WebSocket, request: IncomingMessage, defaultContext: any = {}): void {
     const clientConnection = new ClientConnection(incoming, request, this, this.hooks.bind(this), this.debugger, {
       requiresAuthentication: this.requiresAuthentication,
       timeout: this.configuration.timeout,
-    })
+    }, defaultContext)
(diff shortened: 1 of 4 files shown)


packages/server/src/Hocuspocus.ts
```diff
@@ -372,8 +373,8 @@ export class Hocuspocus {
    * … and if nothings fails it’ll fully establish the connection and
    * load the Document then.
    */
-  handleConnection(incoming: WebSocket, request: IncomingMessage, context: any = null): void {
-    const clientConnection = new ClientConnection(incoming, request, context, this, this.hooks.bind(this), this.debugger, {
+  handleConnection(incoming: WebSocket, request: IncomingMessage): void {
+    const clientConnection = new ClientConnection(incoming, request, this, this.hooks.bind(this), this.debugger, {
```
janthurau, line 377: Can we inject `context` into clientConnection.hookPayload.context? Otherwise this would be a BC change if used in Express or another framework

janthurau (commented):
Thanks again @YousefED! :)

PR #637, opened 2023-06-20
fix: websocketProvider not destroyed

lin52025iq:


janthurau:
this won't work unfortunately because this will close the websocket even if there's another provider still connected via it.

I think that could be done is either add the same logic to `disconnect()` that we already have in `destroy()` (send CLOSE message to the server), and then make sure that the HocuspocusProviderWebsocket will not reconnect after the server closed the connection.

Issue #566, opened 2023-03-30
v2 will not reauthenticate upon re-connect


reknih:
When the `HocuspocusProviderWebsocket` re-establishes the connection to a server with required authentication after some time being offline, its bound `HocuspocusProvider`s will not retransmit their auth tokens because their `isAuthenticated` property is true. Instead, server authentication will fail, and the server will transmit an `Unauthorized` code and close the connection to forbid the client to retry.

**Reproduction:**

- Use a multiplexing-capable version of Hocuspocus server and clients
- Have the client join a server that shall require authentication through an `onAuthenticate` function for all inbound connections
- Cut the connection between client and server
- *Await the server's timeout so that it closes the client socket and voids its authentication*
- Re-Establish the connection between client and server
- A new socket will be created, communication will fail with the error message ``[HocuspocusProvider] An authentication token is required, but you didn’t send one. Try adding a `token` to your HocuspocusProvider configuration. Won’t try again.``

I can reproduce this bug with Node v18.4.0, Chrome 111, and Safari 16.3 on macOS 13.2.1

joshua7v:
I probably hit the same issue here. 
I was trying to figure out why the connect get closed after several seconds and never reconnect back.
And I notice the exact same warning in the console
`[HocuspocusProvider] An authentication token is required, but you didn’t send one.`

I'm a first-day user, and the warning looks quite clear, at first glance at least.
I thought all its complaining is I did not add the token in config as this doc states: https://tiptap.dev/hocuspocus/guides/auth
but seems its not that simple, adding the token does not make any difference
now I'm lost and have no idea how to fix that 😢

janthurau:
hmm, I just tried reproduce this but it just works here. When turning connectivity back on, a new websocket starts with this auth message:

<img width="1255" alt="image" src="https://user-images.githubusercontent.com/896179/231260208-afb3266f-35f4-4936-a621-b87ee70646e8.png">

stedabee:
Any updates on this? We have the same issue here. If we either pass or not token, the warning message occurs, changing something inside the provider, causing a restart in my case. We are using hocuspocus server & provider 2+ .

janthurau:
@stedabee I couldnt reproduce this, but have added another safeguard in v2.0.6. Are you using this version in the provider?

janthurau:
[...]

PR #643, opened 2023-07-03
Add CloseEvents.IdleTimeout

raineorshine:
When the connection becomes idle, it is currently closed with an `Unauthorized` close event. However, this isn't a good fit for an idle timeout. The browser gives an inaccurate warning: `An authentication token is required, but you didn’t send one. Try adding a `token` to your HocuspocusProvider configuration. Won’t try again.`

I added an `IdleTimeout` close event that better matches the reason.

### DIFF
--- a/packages/server/src/ClientConnection.ts
+++ b/packages/server/src/ClientConnection.ts
@@ -1,6 +1,6 @@
 import { IncomingHttpHeaders, IncomingMessage } from 'http'
 import {
-  Forbidden, Unauthorized, WsReadyStates,
+  Forbidden, IdleTimeout, Unauthorized, WsReadyStates,
 } from '@hocuspocus/common'
 import * as decoding from 'lib0/decoding'
 import { v4 as uuid } from 'uuid'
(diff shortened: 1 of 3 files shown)


raineorshine:
Not directly relevant to this PR, but I was wondering if there is a way to restore the connection after it has been closed as idle, or if this happens automatically. Thanks!

packages/server/src/ClientConnection.ts
```diff
@@ -84,7 +84,7 @@ export class ClientConnection {
   ) {
     // Make sure to close an idle connection after a while.
     this.closeIdleConnectionTimeout = setTimeout(() => {
-      websocket.close(Unauthorized.code, Unauthorized.reason)
+      websocket.close(IdleTimeout.code, IdleTimeout.reason)
```
janthurau, line 87: This timeout is stopped once the connection is fully established, so it's actually more a "timed out waiting for authentication". This can usually happen if the server expects an auth token but the client doesn't send one. 

Totally makes sense to add a separate timeout for this, but I think we should keep the old error message for this.

janthurau:
@raineorshine Thanks for your PR!

I'm wondering where did you get this warning in the provider? The server has a few other events in which Unauthorized status codes are sent, so just to make sure that you are actually fixing the cause of your issue with this :)

packages/server/src/ClientConnection.ts
raineorshine: Ah, I see. I may have misunderstood this timeout.

raineorshine:
I get the warning in my console after a period of inactivity, though it does not happen very often, so there must be some other conditions involved. The message is confusing since I am definitely sending a `token` with every `HocuspocusProvider`.

raineorshine:
I see now that the root issue is #566.

Issue #636, opened 2023-06-20
WebsocketProvider not destroyed


lin52025iq:
**Description**
HocuspocusProviderWebsocket is not destroyed when HocuspocusProvider's destroy is called.

**Steps to reproduce the bug**
```js
// init provider
const socket = new HocuspocusProviderWebsocket({
  url: 'ws://127.0.0.1:1234',
})

const provider = new HocuspocusProvider({
  websocketProvider: socket,
  name: 'hocuspocus-demo',
  broadcast: false,
})

// do
provider.destory()

// result: the status of websocket alreadly is pending,
```

**Expected behavior**
HocuspocusProviderWebsocket is destroyed

**Screenshot, video, or GIF**

**Environment?**
 - operating system: all
 - browser: yes
 - mobile/desktop: all
 - Hocuspocus version: 2.1.0

**Additional context**

georeith:
+1 I've been meaning to make a ticket about this, I'm manually destroying the websocket provider at the same time in our application.

janthurau:
hmm yep we're still lacking documentation for this unfortunately. It's intended behavior as the socket can be created without a document connected/connecting.

If you want to close the connection, you need to destroy the websocket provider as well (`socket.disconnect()`). Documentation is in the process ..

romansp:
I agree with the intended behavior described above but I noticed something else that seems unexpected.

I'm reusing single `HocuspocusProviderWebsocket` across multiple `HocuspocusProvider` instances and then manually destroy these when required (`onBeforeUnmount` hook in Vue). But it seems that provider doesn't unsubscribe properly from `websocketProvider` which leads to a memory leak.

For example, `onStatus` hook will still be called for websocket reconnect events later even after `HocuspocusProvider` has been explicitly destroyed.
 
Here handler is configured, but `this.removeAllListeners()` in destroy call won't unsubscribe this.
https://github.com/ueberdosis/hocuspocus/blob/6ca39f2b5768ea1abadd1c88265f5b85d564a3d4/packages/provider/src/HocuspocusProvider.ts#L180

Will try to provide a minimal repro soon.

janthurau:
Closing this, as the behaviour is intended. The memory leak has already been fixed by @raineorshine :)

PR #645, opened 2023-07-06
Unsubscribe provider listeners from websocket on destroy

raineorshine:
In the `HocuspocusProvider` constructor, provider-specific event handlers are added to the `HocuspocusProviderWebsocket` instance, but never removed. This causes a memory leak (probably the one described in https://github.com/ueberdosis/hocuspocus/issues/636#issuecomment-1612237489. When I take a memory snapshot after the provider is destroyed, it shows that the instance is incorrectly retained.

This PR removes all listeners from the `HocuspocusProviderWebsocket` instance when the `HocuspocusProvider` is destroyed. A new memory snapshot shows that the destroyed `HocuspocusProvider` instance is now properly garbage collected.

(`boundBroadcastChannelSubscriber` and `boundBeforeUnload` already existed, and were just moved so all bound methods are together.)

### DIFF
--- a/packages/provider/src/HocuspocusProvider.ts
+++ b/packages/provider/src/HocuspocusProvider.ts
@@ -166,24 +166,24 @@ export class HocuspocusProvider extends EventEmitter {
     this.on('authenticationFailed', this.configuration.onAuthenticationFailed)
 
     this.configuration.websocketProvider.on('connect', this.configuration.onConnect)
-    this.configuration.websocketProvider.on('connect', (e: Event) => this.emit('connect', e))
+    this.configuration.websocketProvider.on('connect', this.forwardConnect)
 
-    this.configuration.websocketProvider.on('open', this.onOpen.bind(this))
-    this.configuration.websocketProvider.on('open', (e: Event) => this.emit('open', e))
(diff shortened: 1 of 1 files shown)


packages/provider/src/HocuspocusProvider.ts
```diff
@@ -414,6 +434,20 @@ export class HocuspocusProvider extends EventEmitter {
 
     this.removeAllListeners()
 
+    this.configuration.websocketProvider.off('connect', this.configuration.onConnect)
```
raineorshine, line 437: My only concern is with unsubscribing from configuration methods. If the configuration is shared among multiple `HocuspocusProvider` instances, the handler will be removed for all of them when only one is destroyed. 

It seems like we could either automatically unsubscribe the configuration method when the configuration is no longer being used, or we could expose a method like `removeConfigurationListeners` and leave it up to the user. Open to your thoughts.

janthurau: Thanks for pointing this out! But indeed, as the configuration already contains the document identifier, it shouldn't be shared at all. Also I don't this we can work around this, so I think it's fine 👍

PR #652, opened 2023-07-14
refs #566: force-reconnect the websocket (if disconnected) on attach of provider

janthurau:
Waiting for feedback in #566, just to make sure.


raineorshine:
This fixes the idle disconnect for me. Thanks!

PR #656, opened 2023-07-20
afterUnloadDocument hook

georeith:
This PR adds an `afterUnloadDocument` hook that runs whenever a previously loaded document is removed by the server, for any reason.

The motivation for adding this feature is because I monitor documents being opened and closed in a load balancer that speaks to an extension and I noticed the load balancer was not removing all documents for some reason when I listen to it like this:

```ts
{
  async afterLoadDocument({
    documentName,
  }: onLoadDocumentPayload): Promise<void> {
   addToLoadBalancer(documentName);
  }

  async onDisconnect({
    documentName,
    clientsCount,
  }: onDisconnectPayload): Promise<void> {
    if (clientsCount === 0) {
      removeFromLoadBalancer(documentName);
    }
  }
}
```

I also fixed up the type of the `hooks` function so that you can no longer pass an incorrect payload that does not match the exact one for that hook specified in the types. There were a couple of cases where they didn't match and I had to fix those. This was to prevent issues such as https://github.com/ueberdosis/hocuspocus/pull/641 in the future.

### DIFF
--- a/packages/server/src/Hocuspocus.ts
+++ b/packages/server/src/Hocuspocus.ts
@@ -19,7 +19,7 @@ import {
   Configuration,
   ConnectionConfiguration,
   HookName,
-  HookPayload,
+  HookPayloadByName,
   beforeBroadcastStatelessPayload,
   onChangePayload,
   onDisconnectPayload,
@@ -398,8 +398,7 @@ export class Hocuspocus {
         }, true)
       } else {
         // Remove document from memory immediately
-        this.documents.delete(document.name)
-        document.destroy()
+        this.unloadDocument(document)
       }
     })
   }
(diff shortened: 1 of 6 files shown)


packages/server/src/Hocuspocus.ts
```diff
@@ -512,7 +511,7 @@ export class Hocuspocus {
       })
     } catch (e) {
       this.closeConnections(documentName)
-      this.documents.delete(documentName)
```
georeith, line 515: I noticed this previously didn't call `document.destroy()` I presume it is safe to do so.

georeith:
@janthurau just bringing to your attention.

janthurau (approved):
Thanks a lot for this!! :-)

PR #658, opened 2023-07-25
fix: remove awareness stats on page unload instead of beforeunload

iowxy:
I notice the code in file `packages/provider/src/HocuspocusProvider.ts`
```typescript
window.addEventListener('beforeunload', this.boundBeforeUnload)
```
I must say that what if I prevent the page to be close? consider this logic below
```typescript
// https://developer.mozilla.org/en/docs/Web/API/Window/beforeunload_event
window.addEventListener('beforeunload', (event) => {
  event.preventDefault()
  event.returnValue = ''
})
```
Develop can write intercept logic and users may cancel to close this page.

### DIFF
--- a/packages/provider/src/HocuspocusProvider.ts
+++ b/packages/provider/src/HocuspocusProvider.ts
@@ -209,7 +209,7 @@ export class HocuspocusProvider extends EventEmitter {
 
   boundBroadcastChannelSubscriber = this.broadcastChannelSubscriber.bind(this)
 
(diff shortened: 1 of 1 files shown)


janthurau (approved):
Totally makes sense. Thanks for this!

I guess we added beforeUnload to make sure the awareness message can actually arrive at the server, but seems this works fine (https://github.com/yjs/y-websocket/issues/119).

PR #669, opened 2023-08-07
throw error in onStoreDocument to prevent afterStoreDocument from releasing lock

ralphiee22:
*Note: This scenario occurs when using the `redis-extension`.*

Upon an initial write, we acquire a redis lock in the `onStoreDocument` hook to prevent multiple writes from colliding with each other.
On subsequent writes, if no error is thrown in `onStoreDocument` hook in `redis-extension` (due to a lock being held already), then `afterStoreDocument` will be called for the server. `afterStoreDocument` will release the lock for that subsequent write, which opens up the flood gates for writes to pile up, essentially DDOSing the datastore.

Code where `error` is always null so it never re-throws.
https://github.com/ueberdosis/hocuspocus/blob/4160dd7b93147787e6524dccd57e9012fc065d0d/packages/server/src/Hocuspocus.ts#L549-L555

### DIFF
--- a/packages/server/src/Hocuspocus.ts
+++ b/packages/server/src/Hocuspocus.ts
@@ -546,14 +546,9 @@ export class Hocuspocus {
 
   storeDocumentHooks(document: Document, hookPayload: onStoreDocumentPayload) {
     this.hooks('onStoreDocument', hookPayload)
-      .catch(error => {
-        if (error?.message) {
-          throw error
-        }
-      })
       .then(() => {
         this.hooks('afterStoreDocument', hookPayload).then(() => {
-        // Remove document from memory.
+          // Remove document from memory.
 
           if (document.getConnectionsCount() > 0) {
             return
@@ -561,6 +556,13 @@ export class Hocuspocus {
 
           this.unloadDocument(document)
         })
+      })
+      .catch(error => {
+        console.error('Caught error during storeDocumentHooks', error)
+
+        if (error?.message) {
+          throw error
+        }
       })
   }
 
--- a/tests/server/onStoreDocument.ts
+++ b/tests/server/onStoreDocument.ts
@@ -252,12 +252,7 @@ test('runs hooks in the given order', async t => {
         triggered.push('four')
       },
       async afterStoreDocument() {
-        t.deepEqual(triggered, [
-          'one',
-          'two',
-        ])
-
-        resolve('done')
+        t.fail() // this shouldnt run
       },
     })
 
@@ -265,6 +260,15 @@ test('runs hooks in the given order', async t => {
       onSynced() {
         provider.configuration.websocketProvider.destroy()
         provider.destroy()
+
+        setTimeout(() => {
+          t.deepEqual(triggered, [
+            'one',
+            'two',
+          ])
+
+          resolve(true)
+        }, 250)
       },
     })
   })
@@ -303,13 +307,7 @@ test('allows to overwrite the order of extension with a priority', async t => {
 
     const server = await newHocuspocus({
       afterStoreDocument: async () => {
-        t.deepEqual(triggered, [
-          'zero',
-          'one',
-          'two',
-        ])
-
-        resolve('done')
+        t.fail()
       },
       extensions: [
         new Running(),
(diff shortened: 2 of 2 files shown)


janthurau:
hmm, I'm not sure I get it.

When you have two instances of Hocuspocus, both connected using redis. The scenario is this:

A tries to write, acquires a lock on redis, and persists to Database.
B now tries to write, fails to acquire a lock on redis, stops onStoreDocument chain, but calls afterStoreDocument. This then won't release the lock (because the lock was acquired by A).
A finished saving, calls afterStoreDocument and releases the lock.

next time, it might be that B is first in acquiring, and then B will be performing the actual save.

Sorry if I'm misreading this, I'm just a bit hesitant to change this as I'm not entirely sure why we chose to only throw if `e.message` exists 🤔

ralphiee22:
I guess this scenario can occur if 2 users separately connect to a document on the same instance. This is why I ran into this issue locally.
I overlooked the part where we have a dict of the locks on the instance, so a user connecting to a different instance won't have this lock.
https://github.com/ueberdosis/hocuspocus/blob/f406003c935051d85b3f6af98b487afb2f6276df/packages/extension-redis/src/Redis.ts#L229


> I'm just a bit hesitant to change this as I'm not entirely sure why we chose to only throw if e.message exists 

right, seems it should throw for any error

ralphiee22:
I think this may also be why in part I see 2 transactions run very closely together for the same document on the same instance (as it doesn't seem to respect the 1 second lock timeout)

long running transaction@2x

georeith:
@janthurau I believe this is a very serious bug. If you have `redis` extension downstream from your DB extension, someone makes a change to a document, disconnects and causes the document to unload from the server.

If this rejects without an error the initial rejection blocks your DB extension from running `onStoreDocument` however this line then catches it:

https://github.com/ueberdosis/hocuspocus/blob/69b97056ed6836142536f564b117c39eb8c2b045/packages/server/src/Hocuspocus.ts#L488-L498

Which allows the unload to proceed. So the document gets discarded on the server without ever saving to your DB.

janthurau:
hmm, I've now spent some time on this. We cannot merge the current PR, as this breaks the extension. If two servers try to write to the DB simultaneously, the 'loser' will crash, which is not what we want.

I think what we might be able to do is just switch `catch` and `then`, so `then` will only execute if onStoreDocument doesn't reject, and catch will make sure that the server doesn't crash.

talhazubairbutt:
> > hmm, I've now spent some time on this. We cannot merge the current PR, as this breaks the extension. If two servers try to write to the DB simultaneously, the 'loser' will crash, which is not what we want.
> > I think what we might be able to do is just switch `catch` and `then`, so `then` will only execute if onStoreDocument doesn't reject, and catch will make sure that the server doesn't crash.
> 
> @janthurau Makes sense in theory to me, the data still has the potential to get stuck on the server but at least it's there if you reconnect. To add some more detail:
> 
> I think I saw this issue because of immediate unload: #683
> 
> Basically I had a pipeline of updates for a single document that came in over a message queue one after the other in quick succession, each one opened and closed the document and caused a full, load, unload cycle on the server which means each one was then queued up with a lock on `onStoreDocument` at the same time.
> 
> Because they were all disconnected at this point they were not exchanging their messages with each other, so the one that eventually won the lock did not have all the updates of all the other event looped `onStoreDocument` waiting for the lock too.
> 
> Eventually the others timed out on retrieving the lock as there were too many waiting for it, and discarded their data, losing it to the void because of the above.
> 
> Without immediate unload this is much harder to make happen. I still think its possible if you have a highly clustered setup and manage to unload the same document in quick succession across the cluster.

@georeith to fix this we use the same lock in onLoad and onDisconnect. This causes onLoad or onDisconnect to retry in case of contention. If onDisconnect retries, it will read the latest value from the servers document list that should have the changes that are applied by the next transaction. if onLoad retries, it will read the value that onDisconnect wrote. If you have a single server setup reading the latest local doc should be fine, if you have multiple servers, you would need to write the doc in redis on both load/disconnect and the latest value on retry.

janthurau (approved):

PR #670, opened 2023-08-09
Always clean up connection listeners

canadaduane:
The Connection class always adds websocket listeners but only conditionally removes them on close. This results in event listener leakage. Instead, we should always remove these listeners when the Connection is closed.

### DIFF
--- a/packages/server/src/Connection.ts
+++ b/packages/server/src/Connection.ts
@@ -163,14 +163,12 @@ export class Connection {
 
       if (this.document.hasConnection(this)) {
         this.document.removeConnection(this)
-        clearInterval(this.pingInterval)
-
-        this.webSocket.removeListener('close', this.boundClose)
-        this.webSocket.removeListener('message', this.boundHandleMessage)
-        this.webSocket.removeListener('pong', this.boundHandlePong)
-
         this.callbacks.onClose.forEach((callback: (arg0: Document, arg1?: CloseEvent) => any) => callback(this.document, event))
       }
+
+      this.webSocket.removeListener('close', this.boundClose)
+      this.webSocket.removeListener('message', this.boundHandleMessage)
+      this.webSocket.removeListener('pong', this.boundHandlePong)
 
       done()
     })


packages/server/src/Connection.ts
```diff
@@ -163,14 +163,13 @@ export class Connection {
...
-        this.webSocket.removeListener('close', this.boundClose)
-        this.webSocket.removeListener('message', this.boundHandleMessage)
-        this.webSocket.removeListener('pong', this.boundHandlePong)
+      this.webSocket.removeListener('close', this.boundClose)
+      this.webSocket.removeListener('message', this.boundHandleMessage)
+      this.webSocket.removeListener('pong', this.boundHandlePong)
 
-        this.callbacks.onClose.forEach((callback: (arg0: Document, arg1?: CloseEvent) => any) => callback(this.document, event))
-      }
+      this.callbacks.onClose.forEach((callback: (arg0: Document, arg1?: CloseEvent) => any) => callback(this.document, event))
```
canadaduane, line 172: There's an open question in our minds about whether this callbacks.onClose is intended to be conditional on the document.hasConnection... we think the callbacks should always fire when the connection is closed?

packages/server/src/Connection.ts
```diff
@@ -163,14 +163,13 @@ export class Connection {
 
       if (this.document.hasConnection(this)) {
         this.document.removeConnection(this)
-        clearInterval(this.pingInterval)
```
canadaduane, line 166: This clearInterval happens 5 lines above, and should not be necessary again.

packages/server/src/Connection.ts
canadaduane: I moved this back inside the condition as it seems required to pass tests; however, I didn't look into it further.

canadaduane:
> I'm getting a similar message, is it because of this?

Yes! That's the same warning message we were seeing.

janthurau:
hmm, not sure if this will fix the warning (as we will still add one listener per document), and I don't see how we can delete connections without matching this condition. I think the warning will only be fixed by https://github.com/ueberdosis/hocuspocus/pull/671 , but still happy to merge this :-) THX!

janthurau (approved):

PR #671, opened 2023-08-09
Linear time mux

canadaduane:
Reduces the amount of work needed to ignore irrelevant messages due to documentName mux/demux from O(n^2) to O(n).

When just a few documents are muxed over the same websocket, the extra work done by the `on('message')` handler in each Connection is not noticable. However, when a single websocket carries messages for dozens or hundreds of documents, the work done by each Connection to discover the documentName becomes ~~exponential~~ quadratic, causing slowdown and potential crashes.

### DIFF
--- a/packages/server/src/ClientConnection.ts
+++ b/packages/server/src/ClientConnection.ts
@@ -29,9 +29,7 @@ import { getParameters } from './util/getParameters.js'
  */
 export class ClientConnection {
   // this map indicates whether a `Connection` instance has already taken over for incoming message for the key (i.e. documentName)
-  private readonly documentConnections: Record<string, boolean> = {}
-
-  private readonly connections: Record<string, Connection> = {}
+  private readonly documentConnections: Record<string, Connection> = {}
 
   // While the connection will be establishing messages will
   // be queued and handled later.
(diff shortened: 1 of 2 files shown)


packages/server/src/ClientConnection.ts
```diff
@@ -31,6 +31,8 @@ export class ClientConnection {
   // this map indicates whether a `Connection` instance has already taken over for incoming message for the key (i.e. documentName)
   private readonly documentConnections: Record<string, boolean> = {}
 
+  private readonly connections: Record<string, Connection> = {}
```
janthurau, line 34: I think we can use `documentConnections` for this functionality (and just change typing from `string, boolean` to `string, Connection`)

packages/server/src/ClientConnection.ts
```diff
@@ -301,6 +307,11 @@ export class ClientConnection {
 
       const documentName = decoding.readVarString(tmpMsg.decoder)
 
+      const instance = this.connections[documentName]
+      if (instance) {
+        instance.handleMessage(data)
```
janthurau, line 312: and then move this to line 316 :)

canadaduane: I think I've made the change you requested here, but I'm unable to see what line 316 is now that I pushed! lol

Please confirm I did what you asked before merging :)

janthurau (approved):

packages/server/src/ClientConnection.ts
janthurau: Thanks! 🙏

PR #672, opened 2023-08-09
Lazily createEncoder for IncomingMessage

canadaduane:
Approximately 2/3 of the initialization time of `new IncomingMessage` is taken by the `createEncoder` call within IncomingMessage's constructor.

The vast majority of IncomingMessage usage is to decode the incoming message first to determine if it is relevant--e.g. to determine what document it belongs to, and what type of message it is.

This change improves performance by avoiding the `createEncoder` call until we know it is actually necessary.

### DIFF
--- a/packages/server/src/IncomingMessage.ts
+++ b/packages/server/src/IncomingMessage.ts
@@ -22,7 +22,9 @@ export class IncomingMessage {
   decoder: Decoder
 
   /**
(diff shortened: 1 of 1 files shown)


janthurau (approved):

janthurau:
Thanks! 🙏 🥳

PR #675, opened 2023-08-10
Allow provider awareness to be disabled

raineorshine:
Fixes https://github.com/ueberdosis/hocuspocus/discussions/673

Awareness provides an additional overhead of computation, memory, and network usage for each provider. This can become significant when there are hundreds of documents that are multiplexed through a single websocket connection.

This PR allows users to opt-out of the provider's default Awareness instance by specifying `awareness: null` in the config.

If awareness is disabled...

- The Awareness instance will not be instantiated and AwarenessMessages will not be sent to the server.
- `setAwarenessField` will throw an error.
- AwarenessMessages from the server will be ignored.

Open to feedback, and will add documentation if approved :)

### DIFF
--- a/packages/provider/src/HocuspocusProvider.ts
+++ b/packages/provider/src/HocuspocusProvider.ts
@@ -59,7 +59,7 @@ export interface CompleteHocuspocusProviderConfiguration {
   /**
    * An Awareness instance to keep the presence state of all clients.
    */
-  awareness: Awareness,
+  awareness: Awareness | null,
   /**
    * A token that’s sent to the backend for authentication purposes.
    */
@@ -106,6 +106,10 @@ export interface CompleteHocuspocusProviderConfiguration {
    * Pass `false` to close the connection manually.
    */
   preserveConnection: boolean,
+}
+
+export class AwarenessError extends Error {
+  code = 1001
 }
 
 export class HocuspocusProvider extends EventEmitter {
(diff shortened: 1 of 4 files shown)


packages/provider/src/HocuspocusProvider.ts
```diff
@@ -108,6 +108,10 @@ export interface CompleteHocuspocusProviderConfiguration {
   preserveConnection: boolean,
 }
 
+export class AwarenessError extends Error {
+  code = 1001
```
raineorshine, line 112: An arbitrary error code that is unique across the app. Even though errors can be tested with `err instanceof AwarenessError`, it's a good idea to have a serializable error code that can be logged, referenced in documentation, and compared across library versions.

packages/provider/src/HocuspocusProviderWebsocket.ts
```diff
@@ -98,8 +98,6 @@ export class HocuspocusProviderWebsocket extends EventEmitter {
     url: '',
     // @ts-ignore
     document: undefined,
-    // @ts-ignore
-    awareness: undefined,
```
raineorshine, line 102: I don't think this is being used anywhere now that the websocket is separate from the provider(s).

packages/provider/src/HocuspocusProviderWebsocket.ts
```diff
@@ -369,7 +367,8 @@ export class HocuspocusProviderWebsocket extends EventEmitter {
     }
 
     // No message received in a long time, not even your own
-    // Awareness updates, which are updated every 15 seconds.
+    // Awareness updates, which are updated every 15 seconds
+    // if awareness is enabled.
```
janthurau, line 371: will the connection actually survive this (checkConnection logic) if there is no awareness message? 🤔 If no update is happening on server/client, and no awareness is received, I think this method will assume the connection is dead?

raineorshine: The connection will auto-close if there are no messages from any provider for the timeout duration. However, the idea is that you only need a single provider to enable Awareness, not hundreds when there are hundreds of documents.

janthurau: makes sense, thx. I'll add an explanation to the parameter.

janthurau (approved):

PR #679, opened 2023-08-13
Subscribe to channels instead of patterns

### LINKED ISSUE #531: redis prefix may cause `MessageReceiver` not working as expected
see code 
```js
private handleIncomingMessage = async (channel: Buffer, pattern: Buffer, data: Buffer) => {
    const channelName = pattern.toString()
    const [_, documentName, identifier] = channelName.split(':')
    const document = this.documents.get(documentName)

    if (identifier === this.configuration.identifier) {
      return
    }

    if (!document) {
      return
    }

    new MessageReceiver(
      new IncomingMessage(data),
      this.logger,
    ).apply(document, undefined, reply => {
      return this.pub.publishBuffer(
        this.pubKey(document.name),
        Buffer.from(reply),
      )
    })
  }
```
note  the line below:

https://github.com/ueberdosis/hocuspocus/blob/476898bfb892227ea9fbd18433e01e9717b9d549/packages/extension-redis/src/Redis.ts#L268

In this line of code, the format of `channelName` is similar to `prefix:documentName:hostIdentity`, where the prefix is allowed to be customized. When the prefix contains ":", it may become the format which like  `prefix::documentName:hostIdentity` , note that there is an extra ":", which will cause the value of this code to not meet expectations, it will cause `MessageReceiver` to never work normally. It is recommended to deal with `prefix`. Remove the ":"

EugeneZ:
This PR accomplishes mainly two things:

First, it removes the need for splitting on the colon (`:`) separator when trying to process Redis messages. This fixes #531, so now you can have a colon in your prefix or document name without issue (this one bit us bad!).

Second, this is a significant performance increase if you are managing a very large number of documents across multiple HP servers. The reason is that subscribing to a pattern [requires Redis to check every subscribed pattern](https://redis.io/commands/publish/) to figure out who will receive the message. So as the number of documents managed by all instances grow, the number of patterns they are subscribed to grow as well, causing `publish` commands to become slower and slower. This PR avoids subscribing to patterns at all, so instead of `n+m` being approximately equal to the number of documents, `n+m` is now approximately equal to the number of HP servers connected to the Redis node(s).

### DIFF
--- a/packages/extension-redis/src/Redis.ts
+++ b/packages/extension-redis/src/Redis.ts
@@ -96,6 +96,8 @@ export class Redis implements Extension {
 
   logger: Debugger
 
+  messagePrefix: Buffer
+
   public constructor(configuration: Partial<Configuration>) {
     this.configuration = {
       ...this.configuration,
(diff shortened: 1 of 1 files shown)


janthurau:
Thanks for this as well :) Will get this merged next week, just want to test some additional things 😎

janthurau (approved):

PR #682, opened 2023-08-17
onChange should update instead of sync

EugeneZ:
When a document changes, the Redis extension performs a full re-sync exchange with every single other server in the cluster. This is way more expensive than just sending an update.

### DIFF
--- a/packages/extension-redis/src/Redis.ts
+++ b/packages/extension-redis/src/Redis.ts
@@ -290,7 +290,12 @@ export class Redis implements Extension {
    * if the ydoc changed, we'll need to inform other Hocuspocus servers about it.
    */
   public async onChange(data: onChangePayload): Promise<any> {
-    return this.publishFirstSyncStep(data.documentName, data.document)
+    const { documentName } = data
+    const updateMessage = new OutgoingMessage(documentName)
+      .createSyncMessage()
+      .writeUpdate(data.update)
+
+    return this.pub.publishBuffer(this.pubKey(documentName), Buffer.from(updateMessage.toUint8Array()))
   }
 
   /**


ralphiee22:
I don't think redis guarantees message delivery which makes me worry about this change.
If messages are dropped then there will be updates that are missed by other clients.
I think that is part of the safe option of sending the entire document each time that if a single update is missed, then on a subsequent sync it will still have the full change set.

ralphiee22:
If we want to change the system to only emit single change events, then we probably need to re-architect to use something like [redis-streams](https://redis.io/docs/data-types/streams/), where we can more reliably confirm that a client receives emitted messages

EugeneZ:
Good point, did not think about that. It seems that doing a full sync only partially helps, though. If there is a problem with receiving messages, it can affect a full sync, too, right? Definitely okay with closing this PR if y'all think it's worth the tradeoff, but it just seems counter to the idea of YJS efficiency.

ralphiee22:
I think @janthurau could speak more to why this implementation was chosen

janthurau:
Thanks for your PR @EugeneZ! Will check this out next week.

In general it seems better to send the update messages, I'm just not sure what happens on newly connected redis clients, or if that still works if an update message gets lost (as raised by the @ralphiee22). An option could be to implement a simple counter field. If the counter-1 doesn't match previously received update event, the server could request a full sync again.

jmoudrik:
Also - if I understand correctly - currently each server reads from DB on first open on its own. Only changing the message to publish the update (== PR) does not guarantee that all hp nodes in the cluster are on the same page about the current doc state - some might have missed previous updates (because no-one was connected to that server before), but before those changes were written to db by the node who did see the change.

RubaXa:
Hi, thank you so much for the excellent tool! I wanted to check if there have been any updates regarding this PR?

The reason I ask is that we have encountered significant freezes when working with large documents (over 2MB), and this PR seems like a potential solution to the issue.

@janthurau I agree with your concerns and would like to understand how you envision the solution with the counter. Could you share how you would suggest implementing it?

I’d appreciate any feedback you may have. Thanks again! ;]

PR #689, opened 2023-08-29
Reset last message received when creating new socket

TalhaASiddiqi:
Currently we dont reset the last message recieved when connecting a new socket, which can potentially cause disconnection on a socket thats in the process of connecting.
If we reset the last message received on new websocket, it will consequently wait for message on this new ws before closing


janthurau (approved):

janthurau:
Thanks!

PR #694, opened 2023-09-04
Fix typos and unnecessary gendered language

12joan:
I noticed an ungainly and unnecessary usage of gendered language on the auth docs page. This PR replaces the gendered construct with the gramatically correct "...apply updates themselves," in addition to fixing some other minor issues on this and other pages.

The main other change this PR makes is to fix incorrect uses of the contractive "it's" in places where the possessive "its" should instead be used.

### DIFF
--- a/docs/guides/auth.md
+++ b/docs/guides/auth.md
@@ -12,7 +12,7 @@ With the `onAuthenticate` hook you can check if a client is authenticated and au
 
 When throwing an error or rejecting the returned Promise, the connection to the client will be terminated (see [server hooks lifecycle](/server/hooks#lifecycle)). If the client is authorized and authenticated you can also return contextual data such as a user id which will be accessible in other hooks. But you don’t need to.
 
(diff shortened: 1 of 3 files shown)


docs/guides/auth.md
```diff
@@ -12,7 +12,7 @@ With the `onAuthenticate` hook you can check if a client is authenticated and au
 
 When throwing an error or rejecting the returned Promise, the connection to the client will be terminated (see [server hooks lifecycle](/server/hooks#lifecycle)). If the client is authorized and authenticated you can also return contextual data such as a user id which will be accessible in other hooks. But you don’t need to.
 
-For more information on the hook and it's payload checkout it's [section](/server/hooks#on-authenticate).
+For more information on the hook and its payload, check out its [section](/server/hooks#on-authenticate).
```
12joan, line 15: "Check out" as a verbal phrase should be two words. "Checkout," outside the context of Git, is typically a noun.

janthurau:
Thank you! 🙏

PR #698, opened 2023-09-07
Fix multiple sockets on reconnection

talhazubairbutt:
Currently websocket listeners are not cleaned up on disconnection and `createWebSocketConnection` is called from multiple sources which results in multiple sockets to be created every time a user comes back online from sleep or there is a disruption in connectivity. 

Steps to reproduce:
1. Open playground (connect to a remote websocket server as it wont occur if you are connecting to 127.0.0.1)
2. Switch wifi from network a to b
3. Initial socket will timeout depending on your `messageReconnectTimeout` value (hearbeat failure).
5. New socket is created
6. Browser calls onClose after about a minute or so, (network tab shows socket duration instead of pending state once close is received). This event triggers a new connection setup.
7. Two connections are now open both have listeners attached and are sending/received data. However, hocuspocus only has reference to one. Keep repeating the process and the number of sockets will keep on increasing and connected sockets go into retry loop and a new socket is created for each close event sent by the browser.

The online event listener also calls `createWebSocketConnection` 

After this change only 1 socket is active at any given time. This PR is just a reference to fix the underlying issue. It can be fixed in multiple ways. I think the event emitter should be bound with the lifecycle of the websocket. Currently even stale websockets emit events that are handled with the latest context.

Ran lint:fix so there are some extra lint changes. 
`attachWebSocketListeners` stores the reference of attached listeners.
`cleanupWebSocket` removes the listeners before resetting reference of the websocket.
`online` event listener has been removed as it leads to a duplicate socket. Once connection has been established, sockets are retried anyway infinite times. If initial connection fails, the retry wrapper takes over and attempts only the number of times that have been set. 
Added an identifier to the socket instance.

Before:

https://github.com/ueberdosis/hocuspocus/assets/49636500/c045b6f9-ac72-4cd8-b5cb-8836d02991f4

After:
image

### DIFF
--- a/packages/provider/src/HocuspocusProviderWebsocket.ts
+++ b/packages/provider/src/HocuspocusProviderWebsocket.ts
@@ -7,6 +7,7 @@ import * as time from 'lib0/time'
 import * as url from 'lib0/url'
 import type { MessageEvent } from 'ws'
 import { Event } from 'ws'
+import { uuidv4 } from 'lib0/random.js'
 import EventEmitter from './EventEmitter.js'
 import { HocuspocusProvider } from './HocuspocusProvider.js'
 import {
@@ -14,6 +15,8 @@ import {
   onAwarenessChangeParameters, onAwarenessUpdateParameters,
   onCloseParameters, onDisconnectParameters, onMessageParameters, onOpenParameters, onOutgoingMessageParameters, onStatusParameters,
 } from './types.js'
+
+export type HocusPocusWebSocket = WebSocket & { identifier: string };
 
 export type HocuspocusProviderWebsocketConfiguration =
   Required<Pick<CompleteHocuspocusProviderWebsocketConfiguration, 'url'>>
@@ -136,7 +139,9 @@ export class HocuspocusProviderWebsocket extends EventEmitter {
 
   subscribedToBroadcastChannel = false
 
-  webSocket: WebSocket | null = null
+  webSocket: HocusPocusWebSocket | null = null
+
+  webSocketHandlers: { [key: string]: any } = {}
 
   shouldConnect = true
 
@@ -152,15 +157,17 @@ export class HocuspocusProviderWebsocket extends EventEmitter {
   }
 
   connectionAttempt: {
-    resolve: (value?: any) => void
-    reject: (reason?: any) => void
+    resolve: (value?: any) => void;
+    reject: (reason?: any) => void;
   } | null = null
 
   constructor(configuration: HocuspocusProviderWebsocketConfiguration) {
     super()
     this.setConfiguration(configuration)
 
-    this.configuration.WebSocketPolyfill = configuration.WebSocketPolyfill ? configuration.WebSocketPolyfill : WebSocket
+    this.configuration.WebSocketPolyfill = configuration.WebSocketPolyfill
+      ? configuration.WebSocketPolyfill
+      : WebSocket
 
     this.on('open', this.configuration.onOpen)
     this.on('open', this.onOpen.bind(this))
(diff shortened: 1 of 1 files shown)


TalhaASiddiqi:
This fails to run in Next JS SSR.

image

EugeneZ:
Thanks for the PR, this is an interesting bug! Why do you need `crypto`? Why not just use an incrementing number?

talhazubairbutt:
> Thanks for the PR, this is an interesting bug! Why do you need `crypto`? Why not just use an incrementing number?
Yup, fixed.

janthurau (approved):

janthurau:
Thanks a lot for the fix & detailed reproduction steps, and even more thanks for the video :)

Merging this 💪

xymfs:
I customized a WebSocketPolyfill for encryption and decryption of websocket data：
```
class WS extends WebSocket{
    constructor(url) {
        super(url)
    }
    set onmessage(f) {
        super.onmessage = (e) => {
            f({
                data: decrypt(e.data)
            })
        }
    }
    send(message) {
        super.send(encrypt(message))
    }
}
```
```
new HocuspocusProvider({
    websocketProvider: new HocuspocusProviderWebsocket({
        WebSocketPolyfill: WS,
    }),
})
```
But I noticed that this change replaced onmessage with addEventListener. How will I hook the returned information?

TalhaASiddiqi:
> I customized a WebSocketPolyfill for encryption and decryption of websocket data：
> 
> ```
> class WS extends WebSocket{
>     constructor(url) {
>         super(url)
>     }
>     set onmessage(f) {
>         super.onmessage = (e) => {
>             f({
>                 data: decrypt(e.data)
>             })
>         }
>     }
>     send(message) {
>         super.send(encrypt(message))
>     }
> }
> ```
> 
> ```
> new HocuspocusProvider({
>     websocketProvider: new HocuspocusProviderWebsocket({
>         WebSocketPolyfill: WS,
>     }),
> })
> ```
> 
> But I noticed that this change replaced onmessage with addEventListener. How will I hook the returned information?

@xymfs you can override addEventListener over here and capture the listener for `message` handler inside it

xymfs:
> > I customized a WebSocketPolyfill for encryption and decryption of websocket data：
> > ```
> > class WS extends WebSocket{
> >     constructor(url) {
> >         super(url)
> >     }
> >     set onmessage(f) {
> >         super.onmessage = (e) => {
> >             f({
> >                 data: decrypt(e.data)
> >             })
> >         }
> >     }
> >     send(message) {
> >         super.send(encrypt(message))
> >     }
> > }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > ```
> > new HocuspocusProvider({
> >     websocketProvider: new HocuspocusProviderWebsocket({
> >         WebSocketPolyfill: WS,
> >     }),
> > })
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > But I noticed that this change replaced onmessage with addEventListener. How will I hook the returned information?
> 
> @xymfs you can override addEventListener over here and capture the listener for `message` handler inside it

Thanks for the tip, it works fine after modification
```
class WS extends WebSocket{
    constructor(url) {
        super(url)
        const originalAddEventListener = this.addEventListener
        this.addEventListener = (type, listener, options) => {
            if (type === 'message') {
                const wrappedListener = (event) => {
                    listener({ ...event, data: decrypt(event.data)})
                }
                originalAddEventListener.call(this, type, wrappedListener, options)
            }else {
                originalAddEventListener.call(this, type, listener, options)
            }
        }
    }
    send(message) {
        super.send(encrypt(message))
    }
}
```

PR #707, opened 2023-09-27
Send Timeout code instead of Auth failed on initial connection timeout

TalhaASiddiqi:
## Description

When a socket connection times out on the server side (i.e. server does not receive token for specified duration) we send 401 code by default. 
The issue with this is that the token may be delayed for a wide variety of reasons, while sending a 401 prevents auto reconnect loop on the frontend by default, leading to either socket never getting connected from frontend, or frontend showing Unauthorized messaging to the user.
Using 408 code ensures that the frontend can keep reconnecting, and only throw 401 when we actually have an invalid token

### DIFF
--- a/packages/provider/src/HocuspocusProviderWebsocket.ts
+++ b/packages/provider/src/HocuspocusProviderWebsocket.ts
@@ -328,8 +328,8 @@ export class HocuspocusProviderWebsocket extends EventEmitter {
 
     Object.keys(handlers).forEach(name => {
       this.webSocket?.removeEventListener(name, handlers[name])
-      delete this.webSocketHandlers[identifier]
     })
+    delete this.webSocketHandlers[identifier]
     this.webSocket.close()
     this.webSocket = null
   }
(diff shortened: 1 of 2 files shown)


adrien-denat-qatalog:
What is needed for this to be merged? Can we help?

TalhaASiddiqi:
> What is needed for this to be merged? Can we help?

@janthurau if there are no concerns about this PR, then can we get it merged?
Otherwise some feedback would be helpful

janthurau:
I thought this could be considered a breaking change, as it changes the behavior and we're not sure if someone relies on this. Maybe its neglectable though - will check it out again in January. Sorry for the delay!

janthurau:
I've checked this again:

* I think we should keep 401 / Unauthorized here, as 401 seems to be the correct status code for "auth required, but not received"
* 403 should be sent for "auth received, but invalid"
* I've also checked the server code, and it seems implemented properly (=as per the upper two bullet points)

I'm anyway preparing a v3 current with the main goal is to clean stuff up in the provider, as we have some weirdness ongoing there due to me trying to keep as much BC as possible from v1 to v2, but that backfired unfortunately with lots of small inconsistencies.

In the meantime, you can probably just resolve the token before creating the provider? Or you pass connect: false and then call .connect() once you actually got the token.

PR #713, opened 2023-10-05
fix: add documentName into broadcast message

xcfox:
Since I upgraded to v2.6.0, the `provider` will throw an error when creating a connection, like the following:
screenshot

This error occurs because the `documentName` parameter is missing when broadcasting the message.
This PR complements the `documentName` parameter for broadcasting.


janthurau (approved):

janthurau:
Thanks! Seems I uncommented broadcasting from the playground some time back, so I haven't noticed this :( Will push the fix out now.

PR #715, opened 2023-10-05
Fix return type of database extension `store` method

haines:
This PR changes the type of the database extension's `store` method from

```typescript
(data: storePayload) => void
```

to 

```typescript
(data: storePayload) => Promise<void>
```

to make it clear that the function is expected to be asynchronous and will be awaited in the `onStoreDocument` hook.

Also, instead of returning the promise directly from the extensions `onStoreDocument`, it is now awaited to ensure that the call stack on any error is correct.

### DIFF
--- a/packages/extension-database/src/Database.ts
+++ b/packages/extension-database/src/Database.ts
@@ -16,7 +16,7 @@ export interface DatabaseConfiguration {
   /**
    * Pass a function to store updates in your database.
    */
-  store: (data: storePayload) => void,
+  store: (data: storePayload) => Promise<void>,
 }
 
 export class Database implements Extension {
@@ -25,7 +25,7 @@ export class Database implements Extension {
    */
   configuration: DatabaseConfiguration = {
     fetch: async () => null,
-    store: async () => null,
+    store: async () => {},
   }
 
   /**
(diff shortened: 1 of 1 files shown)


janthurau (approved):

GermanJablo:
@janthurau and @haines 

Could there be something wrong with this PR? if I do

`store: (data) => { return; },` I get the error: 

> Type 'void' is not assignable to type 'Promise<void>'

and if I do:

`store: async (data) => { return Promise.resolve();},` I get the error:

> Promise-returning function provided to property where a void return was expected. (eslint@typescript-eslint/no-misused-promises)

haines:
Using an async function is correct. I think the linter is complaining because you are returning something unnecessarily. Try this:

```typescript
store: async (data) => { await Promise.resolve(); }
```

Technically for a no-op I think you could just do

```typescript
store: async (data) => { }
```

but the linter will probably complain about that too.

GermanJablo:
Yeah, I tried those two too and a few other combinations. The linter always complains:
```ts
store: async (data) => void {},
store: async (data) => {},
store: async (data) => await Promise.resolve(),
store: async (data) => { await Promise.resolve(); }
store: async (data) => { return Promise.resolve(); }
```
I don't understand what's going on here. Can you reproduce the error if you enable `eslint@typescript-eslint/no-misused-promises`?

haines:
I'm afraid I'm travelling at the moment so I can't check if I can reproduce for a few days.

I would maybe try explicitly providing the return type like

```typescript
store: async (data): Promise<void> => { await Promise.resolve(); }
```

PR #721, opened 2023-10-09
fixes #718: replace arraybuffer by nodebuffer

### LINKED ISSUE #718: Server has websocket.binarytype set to arraybuffer, not nodebuffer
I saw an issue on my logs where the `ws` library was complaining about a type mismatch when emitting messages.   Confirmed that patching the websocket binarytype to `nodebuffer`  fixes the underlying issues.   Basically, since `ws` is both a browser and node library, we're inadvertently opting in to browser behavior on the server and that's no bueno.

See https://github.com/ueberdosis/hocuspocus/blob/f584cb732d045cd615d0babad16503e6a7903312/packages/server/src/Connection.ts#L68  for the line to fix.

I can probably put together a PR next week, but it's Friday night local time and I'm not gonna get to it til after the weekend.

janthurau:


akwirick:
Thanks @janthurau, that'll fix it.

PR #723, opened 2023-10-10
fixes #709: unload document when there is no other connection left af…

### LINKED ISSUE #709: Documents loaded via direct connection not unloaded on disconnect
**Description**
After direct connection to a document is disconnected, the document is not unloaded and stays in memory.

1. Open direct connection to a document
2. Close direct connection
3. Check Hocuspocus state
4. The document is no unloaded

UPD: I tried to force unload document when direct connection is closed, but it leads to a problem when the document is uploaded without persisting the latest change.

**Expected behavior**
It is expected that direct connections behave the same way as remote connection, after disconnect the document should be unloaded if there are no other connections.

**Environment?**
 - Hocuspocus version: 2.5.0

janthurau:
…ter disconnecting a direct connection

fixes #709

### DIFF
--- a/packages/server/src/DirectConnection.ts
+++ b/packages/server/src/DirectConnection.ts
@@ -45,12 +45,24 @@ export class DirectConnection implements DirectConnectionInterface {
     })
   }
 
-  disconnect() {
-    if (this.document && this.document.getConnectionsCount() === 1) {
-      this.instance.unloadDocument(this.document)
-    }
+  async disconnect() {
+    if (this.document) {
 
-    this.document?.removeDirectConnection()
-    this.document = null
+      this.document?.removeDirectConnection()
+
+      await this.instance.storeDocumentHooks(this.document, {
+        clientsCount: this.document.getConnectionsCount(),
+        context: this.context,
+        document: this.document,
+        documentName: this.document.name,
+        instance: this.instance,
+        requestHeaders: {},
+        requestParameters: new URLSearchParams(),
+        socketId: 'server',
+      })
+
+      this.document = null
+    }
   }
+
 }
--- a/packages/server/src/DirectConnection.ts
+++ b/packages/server/src/DirectConnection.ts
@@ -46,6 +46,10 @@ export class DirectConnection implements DirectConnectionInterface {
   }
 
   disconnect() {
+    if (this.document && this.document.getConnectionsCount() === 1) {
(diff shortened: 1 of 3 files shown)


packages/server/src/DirectConnection.ts
```diff
@@ -46,6 +46,10 @@ export class DirectConnection implements DirectConnectionInterface {
   }
 
   disconnect() {
+    if (this.document && this.document.getConnectionsCount() === 1) {
+      this.instance.unloadDocument(this.document)
```
aonnikov, line 50: I might be wrong, and the other change in this pr fizes that, but this won't work when the server is configured with not immediate unload. If the document was recently used by a remote user, this will lead to data loss.

janthurau: hmm, I think this should work.

regardless of whether immediateUnload is used or not, right before actually unloading we're checking again whether connectionCount is 0 (Hocuspocus.ts L470), so if a direct or provider connection connects in the meantime, the doc won't be unloaded.

This line here also make sure to not even trigger the unload in case someone else is connected.

Let me know if you think I forgot something! I've tried to reproduce an issue with a small test, but haven't been able to, so I think we're good 😇

aonnikov: The problem is not with unload itself, I believe the connectionCount check will work as expected. But even if connectionCount is 0, there might be debounced `storeDocumentHooks` if another user just disconnected.

The document will be unloaded without running `storeDocumentHooks`, and will be destroyed. As a result, the unsaved changes will be lost. This happens, though, only if you did not use `transact`  on the direct connection.

It is something that I faced in my project 😄 I had to use empty `transact(() => {})` to force `storeDocumentHooks`.

janthurau: I've also updated the disconnect() method to fix the edge case that you mentioned. I initially wanted to do it this way, but then wanted to avoid calling storeDocumentHooks twice. But definitely better safe than sorry, this should also fix the issue that you mentioned.

aonnikov: Actually, I'm currently use `transact()` with empty arg just to trigger `storeDocumentHooks`, the document is used as a source for a "snapshot". And since I not planned to do any modifications on the source document, I did not use `transact()`. But probably it makes sense to read the doc content inside `transact()` at all.

Thanks for making this change!

PR #727, opened 2023-10-15
Efficiently route messages through HocuspocusProviderWebsocket

### LINKED ISSUE #724: onMessage is inefficiently called on every Doc
**The problem I am facing**
When multiplexing many Docs over a single `HocuspocusProviderWebsocket`, every message is handled by every `HocuspocusProvider`. Every `HocuspocusProvider` reads and writes the documentName in order to see if it is the intended recipient. 

With hundreds of Docs, this results in excessive CPU usage. It can also result in an out-of-memory crash, because as messages pile up, more and more messages are decoded and stored in memory without time for gc to kick in (reproducible).

**The solution I would like**
The main way to alleviate the issue would be to have a single onMessage handler that decodes the documentName and then routes the message to the appropriate `HocuspocusProvider`. This would also improve memory usage by dramatically reducing the number of listeners. 

It is also possible to perform a more efficient documentName comparison (without reading and re-writing the message), but this is minor compared to the listener consolidation. 

None

**Additional context**
I can attempt a PR, but I wanted to post here first and get your input.

raineorshine:
This PR changes the provider's onMessage running time from O(n) to O(1), where n is the number of all attached providers. It does this by adding a Map to the `HocuspocusProviderWebsocket` configuration that allows O(1) lookup of attached providers by document name. It will then manually call onMessage on the appropriate provider and only that provider. 

Fixes #724

### DIFF
--- a/packages/provider/src/HocuspocusProvider.ts
+++ b/packages/provider/src/HocuspocusProvider.ts
@@ -192,8 +192,6 @@ export class HocuspocusProvider extends EventEmitter {
     this.configuration.websocketProvider.on('open', this.boundOnOpen)
     this.configuration.websocketProvider.on('open', this.forwardOpen)
 
-    this.configuration.websocketProvider.on('message', this.boundOnMessage)
-
     this.configuration.websocketProvider.on('close', this.boundOnClose)
     this.configuration.websocketProvider.on('close', this.configuration.onClose)
     this.configuration.websocketProvider.on('close', this.forwardClose)
@@ -233,8 +231,6 @@ export class HocuspocusProvider extends EventEmitter {
   boundPageUnload = this.pageUnload.bind(this)
 
   boundOnOpen = this.onOpen.bind(this)
-
-  boundOnMessage = this.onMessage.bind(this)
 
   boundOnClose = this.onClose.bind(this)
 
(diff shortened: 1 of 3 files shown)


raineorshine:
I still need to test in the playground and my app, but it should be okay. Ended up being a very simple change.

packages/provider/src/HocuspocusProvider.ts
```diff
@@ -447,10 +443,6 @@ export class HocuspocusProvider extends EventEmitter {
 
     const documentName = message.readVarString()
 
-    if (documentName !== this.configuration.name) {
-      return // message is meant for another provider
-    }
-
     message.writeVarString(documentName)
```
raineorshine, line 446: As an aside, I'm not sure why it is required to read the documentName now that this function does not use it. However, removing the read and write breaks the message. I even tried messing around with the encoder and decoder positions but could not get it to work without leaving the read + write as-is. There is clearly a side effect here that I do not understand.

janthurau: yea, that won't work because of the position. You'd need to manually read the first byte, and then move the pointer by the value of it (first byte contains the length of the string). I guess we can keep it like this though, the performance impact should be neglectable.

janthurau:
Thanks a lot for this! Looks great, will do a round of testing as well and then get this merged 😎

PR #747, opened 2023-11-13
implements graceful shutdown

### LINKED ISSUE #711: Graceful shutdown
**The problem I am facing**
I would like to shut the Hocuspocus server down gracefully so that I don't lose state when the server stops.

**The solution I would like**
I would like a `shutdown(): Promise<void>` method on the `Hocuspocus` instance that

1. calls `close` on the underlying HTTP and WebSocket servers, if present
2. closes client connections gracefully
3. waits until all documents are stored and unloaded, the HTTP and WebSocket servers `close` callbacks have been invoked, then resolves the returned promise.

I think this can probably be achieved by combining the existing `closeConnections` method with the `afterUnloadDocument` hook and checking for `getDocumentsCount` being 0. It's a bit of a pain to implement though, so it'd be nice if it came out of the box.

### LINKED ISSUE #738: Documents are not stored before server shut down when `unloadImmediately` is set to false
**Description**
When `unloadImmediately` is set to false, documents are not stored before server shut down.

No `onStoreDocument` called.
No `afterUnloadDocument` called.

1. `HocuspocusServer.configure` set `unloadImmediately` to `false`
3. register `onStoreDocument` hook for logging
4. frontend connect to any one document.
5. use `ctrl + c` shut down server.

**Expected behavior**

Output some logs printed from hooks to prove that the server stored documents before shutting down.

**Environment?**
 - operating system: macOS Sonoma 14.0
 - browser: Google Chrome 118.0.5993.117
 - mobile/desktop: desktop
 - Hocuspocus version: 2.7.1

janthurau:
fixes #711
fixes #738

### DIFF
--- a/packages/server/src/Hocuspocus.ts
+++ b/packages/server/src/Hocuspocus.ts
@@ -185,6 +185,15 @@ export class Hocuspocus {
           port: this.address.port,
         }
 
+        const signalHandler = async () => {
+          await this.destroy()
+          process.exit(0)
+        }
+
+        process.on('SIGINT', signalHandler)
+        process.on('SIGQUIT', signalHandler)
+        process.on('SIGTERM', signalHandler)
+
         try {
           await this.hooks('onListen', onListenPayload)
           resolve(this)
@@ -286,21 +295,30 @@ export class Hocuspocus {
   /**
    * Destroy the server
    */
-  async destroy(): Promise<any> {
-    this.server?.httpServer?.close()
+  destroy(): Promise<any> {
+    return new Promise(async resolve => {
 
-    try {
-      this.server?.webSocketServer?.close()
-      this.server?.webSocketServer?.clients.forEach(client => {
-        client.terminate()
-      })
-    } catch (error) {
-      console.error(error)
-    }
+      this.server?.httpServer?.close()
 
-    this.debugger.flush()
+      try {
 
-    await this.hooks('onDestroy', { instance: this })
+        this.configuration.extensions.push({
+          async afterUnloadDocument({ instance }) {
+            if (instance.getDocumentsCount() === 0) resolve('')
+          },
+        })
+
+        this.server?.webSocketServer?.close()
+        this.closeConnections()
+      } catch (error) {
+        console.error(error)
+      }
+
+      this.debugger.flush()
+
+      await this.hooks('onDestroy', { instance: this })
+    })
+
   }
 
   /**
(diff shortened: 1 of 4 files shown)


packages/server/src/Hocuspocus.ts
```diff
@@ -286,21 +298,30 @@ export class Hocuspocus {
...
-      console.error(error)
-    }
+      this.server?.httpServer?.close()
 
-    this.debugger.flush()
+      try {
+
+        this.configuration.extensions.push({
+          async afterUnloadDocument({ instance }) {
+            if (instance.getDocumentsCount() === 0) resolve('')
```
haines, line 310: I think the promise won't resolve if there were no documents loaded to start with because `afterUnloadDocument` will never be called?

packages/server/src/Hocuspocus.ts
```diff
@@ -286,21 +298,30 @@ export class Hocuspocus {
...
 
-    try {
-      this.server?.webSocketServer?.close()
-      this.server?.webSocketServer?.clients.forEach(client => {
-        client.terminate()
-      })
-    } catch (error) {
-      console.error(error)
-    }
+      this.server?.httpServer?.close()
```
haines, line 304: I think `close` should take a callback and the promise shouldn't resolve until the callback is called.

packages/server/src/Hocuspocus.ts
```diff
@@ -286,21 +298,30 @@ export class Hocuspocus {
...
-    this.debugger.flush()
+      try {
+
+        this.configuration.extensions.push({
+          async afterUnloadDocument({ instance }) {
+            if (instance.getDocumentsCount() === 0) resolve('')
+          },
+        })
+
+        this.server?.webSocketServer?.close()
```
haines, line 314: Same here - the promise shouldn't resolve until the `close` callback is called.

packages/server/src/Hocuspocus.ts
```diff
@@ -286,21 +298,30 @@ export class Hocuspocus {
...
+        this.closeConnections()
+      } catch (error) {
+        console.error(error)
+      }
+
+      this.debugger.flush()
+
+      await this.hooks('onDestroy', { instance: this })
```
haines, line 322: This means the `destroy` method isn't waiting on the `onDestroy` hooks anymore - the `resolve` is called after all the documents are unloaded regardless of whether the `onDestroy` hooks have completed.

janthurau: true, thanks! Fixing in a new PR, as I can't update this one anymore: https://github.com/ueberdosis/hocuspocus/pull/750/files

packages/server/src/Hocuspocus.ts
janthurau: hmm yep, didnt check this before actually. Will note to change this, as the behaviour has been like this forever

Issue #618, opened 2023-05-31
broadcastStateless extend exclude Connection param


groot9325:
**The problem I am facing**
In this scenario, there are two roles: _admin_ and _user_. The admin has the ability to send control events, such as authorizing the user to read or write, directly through HocuspocusProviderWebsocket. However, all users, including the admin role, receive a broadcast event that the admin client don't need to handle. This looks odd and be confusing for the admin client. Ideally, the admin client should be able to ignore this broadcast event automatically.

admin  send stateless  msg:
```
provider.sendStateless(enableUserWrite)
```

server handle onStateless
```
async onStateless(data: onStatelessPayload) {
    //1. handle enableUserWrite event
    ...
   // 2. tell the client (admin),  server has received the message success and replay ack 
    connection.sendStateless(enableUserWriteAck)
   // 3. broadcast enableUserWrite event to user
   document.broadcastStateless(enableUserWrite)
}
```

**The solution I would like**
broadcastStateless extends an exclude connection param that does not send the  event to itself.

```
  /**
   * Broadcast stateless message to all connections except exclude connection
   */
  public broadcastStateless(payload: string, exclude: Connection | null = null): void {
    this.callbacks.beforeBroadcastStateless(this, payload)

    this.getConnections().forEach(connection => {
      if (exclude === connection) {
        return
      }
      connection.sendStateless(payload)
    })
  }
```

**Alternatives I have considered**
admin client ignore the event  manual
**Additional context**

groot9325:
I have discovered an alternative solution that can address this feature.
```javascript
     //broadcast msg to all client exclude sender itself
    async onStateless(data: onStatelessPayload) {
      const { connection, documentName, document, payload } = data
         
            //publish to sibling ws client that in the different node instace which share msg by redis pubsub 
            document.callbacks.beforeBroadcastStateless(document, payload)
            //publish to sibling ws client that share the same node instance
            document.getConnections().filter(conn => conn !== connection).forEach(conn => {
              conn.sendStateless(payload)
            })

            const ackMsg = `{"event":"ackEvent","message":"success"}`
            connection.sendStateless(ackMsg)
    }
```

PR #757, opened 2023-12-11
Adding the possibility of not applying broadcast to specific connections

linspw:
This is an implementation based on @grootgordon suggestion
https://github.com/ueberdosis/hocuspocus/issues/618


janthurau:
Thanks for your PR! The change looks a bit special, as you could also just do it manually and usually the broadcast really should send to everyone.

I think if we can add something like a filter function, it makes it more suitable for other scenarios as well. What do you think?

`public broadcastStateless(payload: string, filter?: (conn: Connection) => boolean): void`

linspw:
Thank you very much for the answer and sorry for my delay, I adjusted the implementation, what you said makes more sense @janthurau

janthurau (approved):

PR #765, opened 2023-12-28
Update hooks.md

websiddu:
onStoreDocument will also be called when connection is destroyed.


janthurau:
not anymore, thanks to #767 :)

Thanks for providing the pull request!

Issue #766, opened 2023-12-28
Unnecessary onStoreDocumentHooks call on last connection close (potential data loss)


ebads67:
**Description**
Currently onStoreDocument hooks are called using the debouncer logic in two cases:

1. On every changes applied to a document (if not coming from redis)
2. On the last connection close

The latter case clears previously scheduled onStoreDocument hooks by the debouncer, and enforces a new onStoreDocument hooks call even if there has not been any changes to the document since the last store. It appears that it is unnecessary to call onStoreDocument hooks if there is no debounced onStoreDocument hook scheduled to be executed. Because all the changes to the document have already been stored.

Even in the case when the last connection is closing and there is a debounced onStoreDocument waiting to be executed, isn't it better to run that instead of running a new onStoreDocument hooks chain?

This has made an issue in our use case where we have two types of connections; connections with write permission and read-only connections. Read-only connections are for the users with only read permissions and are not able to make any changes to the document. So our customized onStoreDocument logic should not be executed for this type of users/connections. But if a read-only user closes the last connection, onStoreDocument is called for this user, and our customized onStoreDocument fails (our downstream system rejects the update operation). This can cause data loss, as any previously debounced onStoreDocument by users with write permission are discarded.

janthurau:
Thanks for raising this, and more thanks for providing the pull request!

Will be released this week.

PR #767, opened 2023-12-28
Avoiding unnecessary onStoreDocument hooks call on last connection close

ebads67:
This is proposal to address the issue #766

### DIFF
--- a/packages/server/src/Hocuspocus.ts
+++ b/packages/server/src/Hocuspocus.ts
@@ -77,7 +77,7 @@ export class Hocuspocus {
 
   debugger = new Debugger()
 
-  debounce = useDebounce()
+  debouncer = useDebounce()
 
   constructor(configuration?: Partial<Configuration>) {
     if (configuration) {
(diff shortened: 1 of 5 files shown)


websiddu:
it would also be better if there is away to call the onStoreDocument hook on demand, e.g if the user exists the page and then reopens the page, immediately they would not get the changes that were not committed yet, they would not see the changes that they just made?

ebads67:
> it would also be better if there is away to call the onStoreDocument hook on demand, e.g if the user exists the page and then reopens the page, immediately they would not get the changes that were not committed yet, they would not see the changes that they just made?

Being able to call onStoreDocument might be a good idea for some scenarios. But I'm not trying to address that in this PR.
Also this PR neither makes nor fixes the potential issue of not getting the latest changes if opening the document right after closing it.

What this PR does is avoiding unnecessary onStoreDocument calls if there is no changes waiting to be saved in the debouncer.

janthurau (approved):

Issue #596, opened 2023-05-03
Request: Do not require 'token' to be set by Provider for OnAuthenticate in Client


jjgmckenzie:
**The problem I am facing**
I have a cookie stored in the request headers, which is consumed by my auth. However, in order to make use of the OnAuthenticate hook (And to not get messages like ``[HocuspocusProvider] An authentication token is required, but you didn’t send one. Try adding a `token` to your HocuspocusProvider configuration. Won’t try again.`` ) , I need to set token to be non nil. A placeholder token like "please_use_auth" works, but I do not see why it is necessary.

**The solution I would like**
Allow token to be optional, either by default or through configuration

**Alternatives I have considered**
Using a placeholder token such as "please_use_auth"

janthurau:
thx for the report! We noticed the same for our cloud (where we allow anonymous access if wanted), and are also using `notoken` as a value for this case. It's on the list, right now the server will just wait for the auth token if you have an authentication hook (`onAuthenticate` defined).

juliankrispel:
I am passing a token (intentionally) and I'm still getting this message :D

valstu:
> I am passing a token (intentionally) and I'm still getting this message :D

Same here

janthurau:
I cannot reproduce this (@valstu @juliankrispel @DenysPoliarush), if you can provide me with a sandbox / replit I'll have a look right away.

valstu:
I'm not 100% sure yet but this could be related to thing that React in strict mode will call all the effects twice on initial render (mount, unmount and mount again). I destroyed the connection when unmounting (using useEffect return) as it is supposed to but it seems that there are two websocket connections after that which somehow breaks things. I guess the hocuspocus instance was destroyed but not the websocket, anyways it was a late night while debugging this so some of the facts has disappeared from my brain :D  But I'll try to see if I can reproduce some sandbox in the coming days.

AliMamed:
> > I am passing a token (intentionally) and I'm still getting this message :D
> 
> Same here

@juliankrispel @valstu hi. Have you managed to investigate on this? I am also passing token, but i am receiving these messages. It doesn't looks to me that there are any multiple connections and I am pretty sure that every time I initiate connection i pass the token.

raineorshine:
@AliMamed It's being tracked in #566. This ticket is for a different issue.

Issue #594, opened 2023-04-29
HocusPocusProvider V2 initiates 2 connections despite setting the connect flag to false


osama-xs:
**Description**
The HocusPocusProvider initiates 2 connections to our YJS backend despite setting the connect flag (from v1) to true. It should set up 1 connection. 

**Steps to reproduce the bug**
1. Go to the examples in the repository https://github.com/BitPhinix/slate-yjs .
2. Upgrade the hocuspocus version on both the client and the server. 
3. Open the client in a browser. And open the Network tab.  
4. You would see 2 connections established to the hocuspocus backend.

**Expected behavior**
A single connection should be formed between the client and the server.

**Screenshot, video, or GIF**
<img width="1728" alt="Screenshot 2023-04-29 at 11 56 52 PM" src="https://user-images.githubusercontent.com/45958974/235325735-20d397d8-3388-4245-b956-8ab1541e4bf3.png">

**Environment?**
 - operating system: MacBook Pro 2021 16 inch Apple M1 Max Ventura 13.3.1
 - browser: Brave
 - mobile/desktop: Desktop
 - Hocuspocus version: 2.0.6

**Additional context**
The problem seems to be in this line where the connect parameter is not passed to the configuration.:
https://github.com/ueberdosis/hocuspocus/blob/ed6441c1d82e614aea388dfb84617c691d18d524/packages/provider/src/HocuspocusProvider.ts#L220

janthurau:
@osamatanveer I've just added an example to the provider documentation. If you want to use multiplexing, you'll have to create the websocket separately from the provider (see https://tiptap.dev/hocuspocus/provider/examples#multiplexing).

osama-xs:
Thanks @janthurau. We're not trying to use multiplexing. It is just that when we connect using `editor.connect()` we see 2 web socket connection opened in the network tab. We still are facing this issue.

aonnikov:
The same issue for me, as a dirty workaround I disconnect the configured websocket provider in addition to the Hocuspocus provider:
```
        provider.configuration.websocketProvider.disconnect()
        provider.disconnect()
```
Looks dirty, but works fine to me.

wangfupeng1988:
> The same issue for me, as a dirty workaround I disconnect the configured websocket provider in addition to the Hocuspocus provider:
> 
> ```
>         provider.configuration.websocketProvider.disconnect()
>         provider.disconnect()
> ```
> 
> Looks dirty, but works fine to me.

[...]

PR #772, opened 2024-01-04
v3 - WIP

janthurau:
goals:

* ⏳  Separate connection / provider more, make it more easy and remove weird hacks that were added to keep BC with v1
* ✅  Separate Hocuspocus server / handler logic
* subdocs support
* fix other breaking issues (#606)
* maybe we can implement reauthentication (not sure what's needed there ; maybe the provider can just re-send the info, but I think the server needs to be adjusted to handle the incoming message type)
* maybe we can finally upgrade ioredis (https://github.com/ueberdosis/hocuspocus/pull/580)
* https://github.com/ueberdosis/hocuspocus/issues/596
* https://github.com/ueberdosis/hocuspocus/issues/594
* fix all issues related to #813 (connection close codes)
* throwing an error in onConfigure should crash the hocuspocus server (i.e. not let it start)
* ... (more to come)


solirpa:
Consider using the api `Y.diffUpdate`, we don't need to load ydoc into memory.

Here is the simulation `syncstep1` and `syncstep2` code:
```
if messageYjsSyncStep1
    cli_sv = decode(cli_msg)
    srv_state = load(name)

    diff = Y.diffUpdate(srv_state, cli_sv);
    send(writeSyncStep2(diff))

    srv_sv = Y.encodeStateVectorFromUpdate(srv_state)
    send(writeFirstSyncStep(srv_sv))

else if messageYjsSyncStep2
    update = decode(msg)
    srv.persist(update)

......

```

PR #778, opened 2024-01-22
Feature/threads

janthurau:


packages/provider/package.json
```diff
@@ -32,6 +32,7 @@
     "@hocuspocus/common": "^2.9.1-rc.0",
     "@lifeomic/attempt": "^3.0.2",
     "lib0": "^0.2.87",
+    "uuid": "^9.0.0",
```
Moiman, line 35: unnecessary dependency

janthurau: 💯  ✅

PR #784, opened 2024-02-01
fix(provider): Keep existing metadata on threads or comments instead of overriding them

bdbch:
The new threads functions on the Tiptap collab provider allow users to update threads or comments - however this will override the whole data object even if only one key is updated.

This PR will allow users to update only single keys.
@janthurau do you think this makes sense or should we solve this another way?


bdbch:
After discussing we're opting to just send the correct data from the client since keys can't be deleted afterwards.

PR #802, opened 2024-03-14
typo on Transformation example

RicoTrevisan:
It was set as "@hocuspocus/extension-transformation", but it should only be "@hocuspocus/transformation"


janthurau:
🙏 thx!

PR #817, opened 2024-04-08
Checking for type of forceSyncInterval

SynteZZZ:
Protection from stupid error when you set forceSyncInterval to true and DDoS your server.

I made such error, wasn't fun.


janthurau (approved):
haha :) Thanks!

PR #822, opened 2024-05-10
Wait for createDocument to be loaded for subsequent createConnections

### LINKED ISSUE #821: Connections to doc are possible while `onLoadDocument` is still in progress
**Description**
The issue arises when multiple client connections come in for the same document, the first connection will kick off and await the `onLoadDocument` hook, however subsequent connections (while its still loading) incorrectly receive a connection to an empty document.  When paired with `@hocuspocus/provider` it changes the meaning on `onConnect` and `onSync`.  For the first client both of those hooks imply that the document has been loaded.  For the second connection it is ambiguous if the document has been loaded.

* client1 connects to doc
* client1 connection invokes `createDocument`
* `this.documents.set(documentName, document)` immediately happens
* client1 starts `onLoadDocument` async hook
* client2 connects, sees that `this.documents.has(documentName)` is true and receives an empty doc, and successfully connects to an empty doc
* client1 `onLoadDocument` hook resolves
* client1 successfully connects

**Expected behavior**
* client1 connects to doc
* client1 connection invokes createDocument
* client1 starts `onLoadDocument` async hook
* client2 connects, sees that `onLoadDocument` is still in progress
* client1 `onLoadDocument` hook resolves
* client1 successfully connects
* client2 successfully connects

jordangarcia:
### Changes

* in `Hocuspocus` add a mapping `loadingDocuments` of `Map<string, Promise<Document>>` to store loading documents
* `createDocument` will use `loadingDocuments` to dedupe docs, always returning a promise of the fully loaded document
* Eliminate tests where it was testing clients connecting to a doc still loading.

Fixes #821

### DIFF
--- a/packages/server/src/Hocuspocus.ts
+++ b/packages/server/src/Hocuspocus.ts
@@ -72,6 +72,8 @@ export class Hocuspocus {
     onDestroy: () => new Promise(r => r(null)),
   }
 
+  loadingDocuments: Map<string, Promise<Document>> = new Map()
+
   documents: Map<string, Document> = new Map()
 
   server?: HocuspocusServer
@@ -414,15 +416,23 @@ export class Hocuspocus {
   /**
    * Create a new document by the given request
    */
-  public async createDocument(documentName: string, request: Partial<Pick<IncomingMessage, 'headers' | 'url'>>, socketId: string, connection: ConnectionConfiguration, context?: any): Promise<Document> {
-    if (this.documents.has(documentName)) {
-      const document = this.documents.get(documentName)
+  public createDocument(documentName: string, request: Partial<Pick<IncomingMessage, 'headers' | 'url'>>, socketId: string, connection: ConnectionConfiguration, context?: any): Promise<Document> {
+    if (this.loadingDocuments.has(documentName)) {
+      const documentPromise = this.loadingDocuments.get(documentName)
 
-      if (document) {
-        return document
+      if (documentPromise) {
+        return documentPromise
       }
     }
 
+    const loadDocPromise = this.loadDocument(documentName, request, socketId, connection, context)
+
+    this.loadingDocuments.set(documentName, loadDocPromise)
+
+    return loadDocPromise
+  }
+
+  async loadDocument(documentName: string, request: Partial<Pick<IncomingMessage, 'headers' | 'url'>>, socketId: string, connection: ConnectionConfiguration, context?: any): Promise<Document> {
     const requestHeaders = request.headers ?? {}
     const requestParameters = getParameters(request)
 
--- a/packages/server/src/Hocuspocus.ts
+++ b/packages/server/src/Hocuspocus.ts
@@ -417,12 +417,9 @@ export class Hocuspocus {
    * Create a new document by the given request
    */
   public createDocument(documentName: string, request: Partial<Pick<IncomingMessage, 'headers' | 'url'>>, socketId: string, connection: ConnectionConfiguration, context?: any): Promise<Document> {
-    if (this.loadingDocuments.has(documentName)) {
-      const documentPromise = this.loadingDocuments.get(documentName)
-
-      if (documentPromise) {
-        return documentPromise
-      }
+    const existingLoadingDoc = this.loadingDocuments.get(documentName)
+    if (existingLoadingDoc) {
+      return existingLoadingDoc
     }
 
     const loadDocPromise = this.loadDocument(documentName, request, socketId, connection, context)
--- a/packages/server/src/Hocuspocus.ts
+++ b/packages/server/src/Hocuspocus.ts
@@ -572,6 +572,7 @@ export class Hocuspocus {
     if (!this.documents.has(documentName)) return
 
     this.documents.delete(documentName)
+    this.loadingDocuments.delete(documentName)
     document.destroy()
     this.hooks('afterUnloadDocument', { instance: this, documentName })
   }
--- a/packages/server/src/Hocuspocus.ts
+++ b/packages/server/src/Hocuspocus.ts
@@ -416,7 +416,7 @@ export class Hocuspocus {
   /**
    * Create a new document by the given request
    */
-  public createDocument(documentName: string, request: Partial<Pick<IncomingMessage, 'headers' | 'url'>>, socketId: string, connection: ConnectionConfiguration, context?: any): Promise<Document> {
+  public async createDocument(documentName: string, request: Partial<Pick<IncomingMessage, 'headers' | 'url'>>, socketId: string, connection: ConnectionConfiguration, context?: any): Promise<Document> {
     const existingLoadingDoc = this.loadingDocuments.get(documentName)
 
     if (existingLoadingDoc) {
@@ -432,9 +432,13 @@ export class Hocuspocus {
 
     this.loadingDocuments.set(documentName, loadDocPromise)
 
-    loadDocPromise.finally(() => {
+    try {
+      await loadDocPromise
       this.loadingDocuments.delete(documentName)
-    })
+    } catch (e) {
+      this.loadingDocuments.delete(documentName)
+      throw e
+    }
 
     return loadDocPromise
   }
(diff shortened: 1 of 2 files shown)


packages/server/src/Hocuspocus.ts
```diff
@@ -414,15 +416,23 @@ export class Hocuspocus {
...
+  public createDocument(documentName: string, request: Partial<Pick<IncomingMessage, 'headers' | 'url'>>, socketId: string, connection: ConnectionConfiguration, context?: any): Promise<Document> {
+    if (this.loadingDocuments.has(documentName)) {
+      const documentPromise = this.loadingDocuments.get(documentName)
 
-      if (document) {
-        return document
+      if (documentPromise) {
+        return documentPromise
       }
     }
```
nperez0111, line 426: This could be simplified
```suggestion
    if (this.loadingDocuments.has(documentName)) {
      return this.loadingDocuments.get(documentName)
    }
 ```

**Review by nperez0111 — APPROVED** [review]
minor nit, but otherwise this looks like a good change to me

File: tests/server/onLoadDocument.ts [hunk 2]
```diff
@@ -164,60 +216,6 @@ test('stops when an error is thrown in onLoadDocument', async t => {
...
-test('disconnects all clients related to the document when an error is thrown in onLoadDocument', async t => {
-  const resolvesNeeded = 4
-
-  await new Promise(async resolve => {
-
-    const server = await newHocuspocus({
-      async onLoadDocument() {
-        return new Promise((resolve, fail) => {
-          setTimeout(() => {
-            // eslint-disable-next-line prefer-promise-reject-errors
```
nperez0111, line 176: Why remove this test?

packages/server/src/Hocuspocus.ts
```diff
@@ -414,15 +416,23 @@ export class Hocuspocus {
...
 
+    const loadDocPromise = this.loadDocument(documentName, request, socketId, connection, context)
+
+    this.loadingDocuments.set(documentName, loadDocPromise)
+
+    return loadDocPromise
+  }
```
nperez0111, line 433: This map will just grow in memory and always short circuit fetching the document (since it will have the promise in memory).
I'd suggest doing a `.finally` to remove this documentName from the map afterwards to clean it up. We don't want this to act as a cache

packages/server/src/Hocuspocus.ts
jordangarcia: This is done for TS / eslint dodging.  `has()` doesn't gaurantee `.get()` will return the value in the map since it could be nullable.  Adding a `!` seems to be prevented by eslint.

tests/server/onLoadDocument.ts
jordangarcia: This test is removed because this behavior should not exist anymore.  This test is asserting that if the `onLoadDocument` throws then `provider2` should still be able to connect to it, which is what we do not want.

packages/server/src/Hocuspocus.ts
jordangarcia: Good catch here on this growing infinitely!  

However, I dont quite follow using `.finally` here.  The issue that I am thinking of is that we use `loadingDocuments` as the state of whether a doc has been loaded in the form of a promise. The only time it seems acceptable to delete from this map is when we call `unloadDocument`.  I just pushed a change that removes the `Promise<Document>` when `unloadDocument` is called.

packages/server/src/Hocuspocus.ts
nperez0111: Then maybe this might suit better

```js
const loadingDocument = this.loadingDocuments.get(documentName);

if (loadingDocument) {
  return loadingDocument
}
```

tests/server/onLoadDocument.ts
nperez0111: But the error may be transient, so if it throws it should be able to retry, no?

packages/server/src/Hocuspocus.ts
nperez0111: @jordangarcia I'm unsure that your change got pushed up here, can you check that?

I think that we are thinking of this mapping a bit differently. I am thinking of this map only being useful for "re-using" a fetch to the loading document. This would make it so that two clients requesting the same document while loading would wait for the same document to finish loading.

Client A fetches Doc 1
Client A waits for Doc 1 to load
Client B fetches Doc 1
Client B waits on the same promise as Client A
Doc loads
Client A syncs
Client B syncs


Whereas if I'm understanding you correctly, you want to store the promise here and use this as the source of truth for documents that are loaded or not. When this is what the `documents` map is meant to do.

Please let me know your thoughts

tests/server/onLoadDocument.ts
jordangarcia: I'm not quite sure i follow.  Wouldn't it be up to the `HocuspocusProvider` to handle retry logic once the connection fails to be established.

packages/server/src/Hocuspocus.ts
jordangarcia: My interpretation of the code is that the `this.documents` map isn't quite intended to keep track of "loaded" documents.  Since we insert into that map before the document is loaded.  `this.documents` seems to represent the documents that this hocuspocus instance is managing, regardless of their loading state.  Changing `this.documents` to represent documents to be loaded seems like a much bigger API change since many functions rely on this mapping to understand the state of the system.
 
The `loadDocuments` map would be used as the state of the loading documents and a way for `createDocument` to guarantee a singular doc always being returned (or a promise to the resolved value).  The fact that `loadDocuments` sets the created document in `this.documents` then returns the document gaurantees referential equality between the two maps.    I do think this may seem a bit fragile to code changes that wouldn't keep these two in sync.

The key issue here is that `createDocument` was using the `this.documents` map as a cache of the documents.  To be consistent we always want `createDocument` to be `Promise<Document>` of the loaded document.  I dont really see a way around not using `loadingDocuments` as a cache here for `createDocument`.  We could add more state to the system and keep track of `Map<string, boolean>` for whether a document is loaded and then use `this.documents` instead of `this.loadingDocuments` but that just seems more complex to me.

tests/server/onLoadDocument.ts
jordangarcia: If I add the following console logs to the test i see:

```
test.only('disconnects all clients related to the document when an error is thrown in onLoadDocument', async t => {
  const resolvesNeeded = 4

  await new Promise(async resolve => {

    const server = await newHocuspocus({
      async onLoadDocument() {
        return new Promise((resolve, fail) => {
          setTimeout(() => {
            // eslint-disable-next-line prefer-promise-reject-errors
            fail('ERROR')
          }, 250)
        })
      },
      async onStoreDocument(data) {
        t.fail('MUST NOT call onStoreDocument')
      },
    })

    let resolvedNumber = 0
    const resolver = () => {
      resolvedNumber += 1

      if (resolvedNumber >= resolvesNeeded) {
        t.is(server.documents.size, 0)
        t.is(server.getConnectionsCount(), 0)
        resolve('done')
      }
    }

    const provider1 = newHocuspocusProvider(server, {
      onConnect() {
        console.log('provider1 connect')
        resolver()
      },
      onClose(event) {
        provider1.disconnect()
        console.log('provider1 disconnect')
        resolver()
      },
    })

    const provider2 = newHocuspocusProvider(server, {
      onConnect() {
        console.log('provider2 connect')
        resolver()
      },
      onClose() {
        provider2.disconnect()
        console.log('provider2 disconnect')
        resolver()
      },
    })

  })

})
```


```
provider2 connect
provider1 disconnect
provider2 disconnect
provider2 disconnect
```

I feel like this is not what i would expect this behavior to do in the first place.

packages/server/src/Hocuspocus.ts
jamesopti: Yea this makes sense to me as well. `createDocument` is idempotent and will always return a Promise that resolves to the singular (referential) source of truth `Document`.

I do see how this technically is adding more "redundant" state, but I think it's quite clear and simple.

@nperez0111 Curious to hear your thoughts!

jamesopti: I think this is a good solution as well. Thanks for helping out!

Would love to get this out as soon as possible.

jamesopti (approved):

Issue #657, opened 2023-07-21
Replace unload with visibility API


georeith:
**Description**

In chrome when clicking on a link that has headers of `Content-Disposition: Attachment` it triggers a file download, if that URL is on an external URL and you can't use the `download` attribute (which only works on same origin) then it also fires the `unload` event on window, despite not actually unloading the page.

HocuspocusProvider listens to `unload` and clears out the awareness state, there is then no follow up event to reinstate it: https://github.com/ueberdosis/hocuspocus/blob/d5cf4212e3d33a9d54909e3082a3c57307bf80a4/packages/provider/src/HocuspocusProvider.ts#L282

Instead we should use the [Page Visibility API](https://developer.mozilla.org/en-US/docs/Web/API/Page_Visibility_API) to clear out awareness (which will work on tabbing in and out also) but also does not trigger when clicking on downloads in Chrome.

Whilst this is partly a quirk of the browser there is a somewhat simple workaround.

**Steps to reproduce the bug**

1. Click on a link to an external file with the header `Content-Disposition: Attachment`
2. See that your awareness state is emptied despite not leaving the page
3. Awareness will not recover despite still being connected to the server

**Expected behavior**

Awareness should not get cleared when clicking on links that don't leave the page.

**Screenshot, video, or GIF**

https://github.com/ueberdosis/hocuspocus/assets/3090066/ad08e5fa-4c38-4a8b-bc56-00e98073ef20

**Environment?**
 - operating system: MacOS 13
 - browser: Chrome 114
 - Hocuspocus version: 2.3.0

georeith:
Actually this issue is fixed by: https://github.com/ueberdosis/hocuspocus/pull/658

janthurau:
😎  closing this - let me know if is is really fixed in 2.3.0 :-)

PR #824, opened 2024-05-23
Replace deprecated `unload` event with `pagehide` event

jul13579:
Dear hocuspocus folks,

I noticed that the `HocuspocusProvider` is using the deprecated `unload` event. MDN advises to use the `pagehide` event instead, so this is what I chose to use for this PR. I searched for any open and closed issues and PRs that could be related to this and hence am aware of #658 and #657, where the previously used `beforeunload` event was found to be problematic.

I checked if the `pagehide` event also fires when starting a download as described in #657 and found that it in fact does not fire. Therefore I think this event might be a viable alternative to `unload`.

Cheers,
Julian


janthurau:
Thanks! Sounds good :)

PR #830, opened 2024-06-12
Allow token fn to return null

fuadsaud:
The token passed as a value can be null already, so the code that calls the function will continue to work as intended.

Allowing a null value makes it possible to still open a connection without a token while attempting to fetch the token in an async function.


fuadsaud:
@janthurau sorry about that, I think prettier kicked in when I commited the file and I didn't double check the diff after that.

I made a [new PR](https://github.com/ueberdosis/hocuspocus/pull/838) 🙏 .

PR #831, opened 2024-06-12
redis memory leaks

janthurau:


packages/extension-redis/src/Redis.ts
```diff
@@ -364,5 +364,6 @@ export class Redis implements Extension {
     await this.redlock.quit()
     this.pub.disconnect(false)
     this.sub.disconnect(false)
+    this.sub.off('messageBuffer', this.handleIncomingMessage)
```
janthurau, line 367: this probably won't help at all, as onDestroy is called when the server shuts down.

packages/server/src/Document.ts
```diff
@@ -260,6 +260,7 @@ export class Document extends Doc {
 
   destroy() {
     super.destroy()
+    this.awareness.destroy()
```
janthurau, line 263: this might fix pending _checkInterval timers

janthurau: ok probably not, as y-protocols/awareness destroys itself when the attached doc is destroyed.

packages/extension-redis/src/Redis.ts
```diff
@@ -134,7 +134,9 @@ export class Redis implements Extension {
     }
     this.sub.on('messageBuffer', this.handleIncomingMessage)
 
-    this.redlock = new Redlock([this.pub])
+    this.redlock = new Redlock([this.pub], {
+      retryCount: 0,
```
janthurau, line 138: not sure why we havent had this, I actually thought 'lockTimeout' would trigger an error if it can't lock after X ; However, what we're doing is just silently ignoring the lock after X...

PR #834, opened 2024-06-13
fixes #832

### LINKED ISSUE #832: DirectConnection causing document state corruption
**Description**

If you use a `DirectConnection` when there are no WebSocket connections for a document, the WebSocket connections established during the next 2 seconds (or debounce time) will end up using an orphaned document and will not sync changes properly with clients that establish connections after that 2 second period. The problem lies in that `storeDocumentHooks` is called directly after closing the DirectConnection, but also in a debounced manner a bit later. Details follow.

1. Start from a situation where there are no connections for document `doc1`.
2. Establish a direct connection on the server, transact something and disconnect the connection. 
3. At this stage [`storeDocumentHooks`](https://github.com/ueberdosis/hocuspocus/blob/879dd266f54bdf04145915c19f0d301da9df1582/packages/server/src/Hocuspocus.ts#L524) has now beed called twice (once for the ([`transact`](https://github.com/ueberdosis/hocuspocus/blob/879dd266f54bdf04145915c19f0d301da9df1582/packages/server/src/types.ts#L376) and once for the [`disconnect`](https://github.com/ueberdosis/hocuspocus/blob/879dd266f54bdf04145915c19f0d301da9df1582/packages/server/src/types.ts#L377)). As the result of the latter, the [`unloadDocument`](https://github.com/ueberdosis/hocuspocus/blob/879dd266f54bdf04145915c19f0d301da9df1582/packages/server/src/Hocuspocus.ts#L581) method has been called and the `doc1` has beed removed from `this.documents`
4. All fine, except that there's still a [debounced](https://github.com/ueberdosis/hocuspocus/blob/879dd266f54bdf04145915c19f0d301da9df1582/packages/server/src/Hocuspocus.ts#L408) call for `storeDocumentHooks` to be performed in 2 seconds or so.
5. Let's say a WebSocket client connects at this stage. Because the `doc1` is not loaded yet, it will now be loaded
6. After 2 seconds, the debounced `storeDocumentHooks` call kicks in. This call is bound to the earlier `Y.Doc` instance that no longer is referenced in `this.documents` (there's no a new copy of `doc1` loaded). This call results into another call to `unloadDocument` which, [on this line](https://github.com/ueberdosis/hocuspocus/blob/879dd266f54bdf04145915c19f0d301da9df1582/packages/server/src/Hocuspocus.ts#L585) removed the document from `this.documents`. Unfortunately, it ends up removing the _new document instance_, which is now referenced by the WebSocket client which connected in step 5.
7. Now the WebSocket client is connected to an orphaned document not in `this.documents`. Any client connecting from now on, will be assigned to a new document instance and the clients will not see changes from each other.

**Expected behavior**

When the DirectConnection is closed and there are no more clients for `doc1`, we should not unload the document when there is still a debounced storeDocumentHooks call left to be performed. Or alternatively, at least make sure that the debounced call does not end up removing the newer, active document from `this.documents`.

In my project, I made a quick patch for [this line](https://github.com/ueberdosis/hocuspocus/blob/879dd266f54bdf04145915c19f0d301da9df1582/packages/server/src/Hocuspocus.ts#L583) to additionally check that the document currently in `this.docs` is actually same document that's now being unloaded. 

The patch solution is suboptimal though at least in the sense that two copies of the same document will exist in server memory for a time. I'd prolly rather have it so that the document accessed by DirectConnection would be subject to the same lifecycle as with WebSocket connections - that is, to be unloaded only after the normal debounce period. Yet, I'm not intimately familiar with the server design, so my hunch may be misguided.

janthurau:
fixes #832

### DIFF
--- a/tests/server/openDirectConnection.ts
+++ b/tests/server/openDirectConnection.ts
@@ -218,3 +218,49 @@ test('direct connection transact awaits until onStoreDocument has finished, even
     resolve('done')
   })
 })
+
+test('creating a websocket connection after transact but before debounce interval doesnt create different docs', async t => {
+  let onStoreDocumentFinished = false
+  let disconnected = false
+
+  await new Promise(async resolve => {
+    const server = await newHocuspocus({
+      onStoreDocument: async () => {
+        onStoreDocumentFinished = false
+        await sleep(200)
+        onStoreDocumentFinished = true
+      },
+      async afterUnloadDocument(data) {
+        console.log('called')
+        if (disconnected) {
+          t.fail('must not be called')
+        }
+      },
+    })
+
+    const direct = await server.openDirectConnection('hocuspocus-test')
+    t.is(server.getDocumentsCount(), 1)
+    t.is(server.getConnectionsCount(), 1)
+
+    t.is(onStoreDocumentFinished, false)
+    await direct.transact(document => {
+      document.transact(() => {
+        document.getArray('test').insert(0, ['value'])
+      }, 'testOrigin')
+    })
+    t.is(onStoreDocumentFinished, true)
+
+    await direct.disconnect()
+    disconnected = true
+
+    t.is(server.getConnectionsCount(), 0)
+    t.is(server.getDocumentsCount(), 0)
+    t.is(onStoreDocumentFinished, true)
+
+    const provider = newHocuspocusProvider(server)
+
+    await sleep(server.configuration.debounce * 2)
+
+    resolve('done')
+  })
+})
--- a/packages/server/src/DirectConnection.ts
+++ b/packages/server/src/DirectConnection.ts
@@ -42,7 +42,7 @@ export class DirectConnection implements DirectConnectionInterface {
       requestParameters: new URLSearchParams(),
       socketId: 'server',
       transactionOrigin,
-    })
+    }, true)
   }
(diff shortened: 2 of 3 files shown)


tests/server/openDirectConnection.ts
```diff
@@ -218,3 +218,49 @@ test('direct connection transact awaits until onStoreDocument has finished, even
     resolve('done')
   })
 })
+
+test('creating a websocket connection after transact but before debounce interval doesnt create different docs', async t => {
+  let onStoreDocumentFinished = false
+  let disconnected = false
+
+  await new Promise(async resolve => {
```
raimohanska, line 226: This Promise wrapper here looks unnecessary to me, as it's used inside an async function anyway and ultimately wraps another async function.

raimohanska: Otherwise, this test case seems to capture the use case 👍

raimohanska: Run the test against the unfixed `main` version and it failed, so it does indeed capture the error and the fix. Nice!

raimohanska:
Looks good to me. Now there's essentially a single path to the execution of storeDocumentHooks and there should be no orphaned document instances. 👍

PR #838, opened 2024-06-24
Allow token fn to return null

fuadsaud:
The token passed as a value can be null already, so the code that calls the function will continue to work as intended.

Allowing a null value makes it possible to still open a connection without a token while attempting to fetch the token in an async function.


fuadsaud:
It seems that this might not work since [this line](https://github.com/ueberdosis/hocuspocus/blob/7b42fa200ab497148fa7c54700588688f4f5416d/packages/provider/src/HocuspocusProvider.ts#L370C1-L372C4) doesn't deal well with a null-returning function. I'll think a bit more about the design.

PR #843, opened 2024-07-05
feat: typed context

jpoehnelt:
This allows passing a generic to `Hocuspocus`:

```ts
type Context = {
foo: string;
}

const server =

### DIFF
--- a/packages/server/src/Hocuspocus.ts
+++ b/packages/server/src/Hocuspocus.ts
@@ -49,7 +49,7 @@ export const defaultConfiguration = {
 /**
  * Hocuspocus Server
  */
-export class Hocuspocus {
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: packages/server/src/Hocuspocus.ts [hunk 1]
```diff
@@ -340,7 +340,7 @@ export class Hocuspocus {
    * … and if nothing fails it’ll fully establish the connection and
    * load the Document then.
    */
-  handleConnection(incoming: WebSocket, request: IncomingMessage, defaultContext: any = {}): void {
+  handleConnection(incoming: WebSocket, request: IncomingMessage, defaultContext: Ctx = {} as Ctx): void {
```
jpoehnelt, line 343: kind of a hack, my typescript-fu is lacking to know what else to do

PR #844, opened 2024-07-08
Don't close socket if preserveConnection is set in HocuspocusProvider

douira:
Sending the `CloseMessage` also has the effect of closing the socket according to its description.

I noticed that before patching this in the package, the socket would be closed when I called `destroy` on the provider even if it was supposed to preserve the connection. When using multiple providers on the same socket at once, it's important that no provider getting destroyed closes the whole socket. `disconnect`ing the provider is not enough because it still has listeners on the document, socket, and elsewhere which is not acceptable.


douira:
The test could be fixed by either
- setting `preserveConnection` to false in the test (it's true by default),
- or changing the default to false,
- or changing how many open connections the test expects,
- or adding a `socket.destroy()` to the test

douira:
Is there still interest in this? I'm still maintaining a `patch-package` hack for this in my project.

janthurau:
Thanks @douira! I'm currently working on v3 and a fix for this will be included.

Issue #854, opened 2024-09-17
extension-redis memory leak with frequent doc updates 


entropi:
**Description**
When performing a large number of concurrent DirectConnection updates to a handful of documents, a rapidly escalating memory leak was observed. Waiting and forcing garbage collection had no impact on memory growth.

This appears to be separate from https://github.com/ueberdosis/hocuspocus/issues/846 as the behavior was observed even when executing rapid updates on only a small number of distinct documents, so it doesn't just appear to be due to not unloading docs.

**Steps to reproduce the bug**

Perform many (hundreds to thousands) of updates (in my case via DirectConnection) to a finite set of documents in a very small window. Observe rapid memory growth, which calls to GC do not clear.

**Expected behavior**
Memory stays at a reasonable level. setTimeout delays to account for sync remain at one per documentName.

**Environment?**
 - Hocuspocus version: 2.13.5

**Additional context**
Lowering the disconnectDelay to 1ms removed the leak (though obviously at the cost of the benefit of the delay for syncs), and raising it to 5000ms during rapid updates escalated the memory growth curve.  I suspect that it's related to the fact that we're potentially stacking hundreds or thousands of setTImeout operations on the same document in afterStoreDocument within the disconnectDelay window during rapid edits.

~A forthcoming~ The Linked PR appears to mitigate the issue by clearing any pending timeout operation on a document when a new afterStoreDocument or onDisconnect hook fires for the same documentName, instead of always adding new ones on top.

janthurau:
Thanks for your work! I've released this update as 2.13.7 :-)

PR #855, opened 2024-09-17
Prevent setTimeout buildup during frequent document updates

entropi:
Addresses https://github.com/ueberdosis/hocuspocus/issues/854

When making many edits to the same document within the disconnectTimeout window (1000ms by default), it's possible to stack a large number of redundant setTimeout calls. This tracks and clears any pending Timeouts when new hooks are fired against the same document.

in the case of afterStoreDocument, we immediately resolve the promise so that outstanding hooks complete. In the case of onDisconnect, we simply clear the timeout, since we'll set a new delayed disconnect call.


entropi:
@janthurau any rough idea when this might be able to be reviewed? it's currently a blocking leak for us, but would prefer not to deploy a fork for too long/at all if avoidable.

PR #865, opened 2024-10-24
Transformer: move @tiptap/core to peerDependencies

trulex:
Make tiptap core peer dependency to prevent discrepancy between version used by transformer and version used by project that relies on transformer.

### DIFF
--- a/packages/transformer/package.json
+++ b/packages/transformer/package.json
@@ -28,16 +28,16 @@
     "dist"
   ],
   "dependencies": {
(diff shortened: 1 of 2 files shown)


packages/transformer/package.json
nperez0111, line 1: Shouldn't these deps just have a looser version range like `^2`?

trulex: That would probably be fine. We'd still need to keep `@tiptap/core` in `peerDependencies` to prevent version mismatch I believe

PR #868, opened 2024-10-28
fix(provider): correct event handler binding for document updates and awareness

RubaXa:
Fixed an issue where event handlers for document and awareness updates leaks on destroy.


nperez0111 (dismissed):
Thanks for the contribution, this looks right to me

nperez0111 (approved):

PR #885, opened 2024-12-09
feature(provider): allow thread tombstoning and restoration

bdbch:
This pull request to `packages/provider` introduces several enhancements and new features related to thread management in the `TiptapCollabProvider` class. The changes include the addition of new options for handling threads, improvements to thread deletion and restoration, and updates to the type definitions.

Enhancements to thread management:

* [`packages/provider/src/TiptapCollabProvider.ts`](diffhunk://#diff-811a6faec592b48056e4a70d311dd8596878f455baf0978454be672cbb8718fcR12-R13): Added new options `DeleteThreadOptions` and `GetThreadsOptions` to control the output and deletion of threads. The `getThreads` method now accepts an `options` parameter to filter threads based on their deletion status. [[1]](diffhunk://#diff-811a6faec592b48056e4a70d311dd8596878f455baf0978454be672cbb8718fcR12-R13) [[2]](diffhunk://#diff-811a6faec592b48056e4a70d311dd8596878f455baf0978454be672cbb8718fcR142-R164)
* [`packages/provider/src/TiptapCollabProvider.ts`](diffhunk://#diff-811a6faec592b48056e4a70d311dd8596878f455baf0978454be672cbb8718fcL245-R326): Introduced the `restoreThread` method to restore deleted threads by setting the `deletedAt` property to `null`.
* [`packages/provider/src/TiptapCollabProvider.ts`](diffhunk://#diff-811a6faec592b48056e4a70d311dd8596878f455baf0978454be672cbb8718fcL245-R326): Updated the `deleteThread` method to support soft deletion (marking as deleted) and forceful deletion of threads and their comments based on the provided options.

Type definition updates:

* [`packages/provider/src/types.ts`](diffhunk://#diff-9672b197f8befe1ede0bdc9d6458f6f076be4029d5951ba8bae18d8565365f8fR113): Added `deletedAt` property to the `TCollabThread` type to indicate the deletion status of a thread.
* [`packages/provider/src/types.ts`](diffhunk://#diff-9672b197f8befe1ede0bdc9d6458f6f076be4029d5951ba8bae18d8565365f8fR201-R226): Defined new types `DeleteThreadOptions` and `GetThreadsOptions` to support the new options in thread management methods.

### DIFF
--- a/packages/provider/src/TiptapCollabProvider.ts
+++ b/packages/provider/src/TiptapCollabProvider.ts
@@ -9,12 +9,18 @@ import {
 import { TiptapCollabProviderWebsocket } from './TiptapCollabProviderWebsocket.js'
 import type {
   DeleteCommentOptions,
+  DeleteThreadOptions,
   TCollabComment, TCollabThread, THistoryVersion,
 } from './types.js'
 
 const defaultDeleteCommentOptions: DeleteCommentOptions = {
   deleteContent: false,
   deleteThread: false,
+}
+
+const defaultDeleteThreadOptions: DeleteThreadOptions = {
+  deleteComments: false,
+  force: false,
 }
 
 export type TiptapCollabProviderConfiguration =
@@ -199,6 +205,7 @@ export class TiptapCollabProvider extends HocuspocusProvider {
       thread.set('createdAt', (new Date()).toISOString())
       thread.set('comments', new Y.Array())
       thread.set('deletedComments', new Y.Array())
+      thread.set('deletedAt', null)
 
       this.getYThreads().push([thread])
       createdThread = this.updateThread(String(thread.get('id')), data)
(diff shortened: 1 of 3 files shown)


packages/provider/src/TiptapCollabProvider.ts
```diff
@@ -17,6 +19,15 @@ const defaultDeleteCommentOptions: DeleteCommentOptions = {
   deleteThread: false,
 }
 
+const defaultGetThreadsOptions: GetThreadsOptions = {
+  types: ['undeleted'],
```
nperez0111, line 23: undeleted isn't a word and doesn't sound right.

How about archived & not archived. Don't need two states, when people get, it should only ever be not archived. And asking for archived means you want archived too.

Like `includeArchived?: true`

nperez0111: Or includeDeleted if you wanna stick to that naming

nperez0111 (approved):
Other than the naming, the change looks good.

packages/provider/src/types.ts
```diff
@@ -217,10 +217,22 @@ export type DeleteThreadOptions = {
   force?: boolean,
 }
 
+export enum ThreadType {
```
nperez0111, line 220: Enums are generally discouraged in TS (just one [example of why](https://www.aaron-powell.com/posts/2020-05-27-the-dangers-of-typescript-enums/))

So typically, a const object would be used instead which mostly has the same properties

PR #888, opened 2024-12-17
Doc: add missing await in "Editing a document locally" example.

blep:
DirectConnection.disconnect() is an async function and should be await to wait for its promise to complete.

Notes: experienced C++/Java dev, but total Typescript beginner, so make sure to check.


janthurau (approved):
Thanks!

Issue #846, opened 2024-07-17
Memory leak in direct connection


yurtsiv:
**Description**

The document is not unloaded after invoking [DirectConnection.disconnect()](https://github.com/ueberdosis/hocuspocus/blob/main/packages/server/src/DirectConnection.ts#L48) if one of [store document hooks](https://github.com/ueberdosis/hocuspocus/blob/edd7034aece9ba41dcb4efff1207fd06624f319d/packages/server/src/Hocuspocus.ts#L519) fails.
The most common scenario is Redis extension failing to [acquire the lock](https://github.com/ueberdosis/hocuspocus/blob/edd7034aece9ba41dcb4efff1207fd06624f319d/packages/extension-redis/src/Redis.ts#L236).

**Scenario**

Server A opens/closes many direct connections to documents without any other connections on that server. However, some documents might be loaded and updated on server B, occasionally causing the aforementioned lock error. As a result server A slowly accumulates loaded documents.

As a workaround, I'm periodically unloading documents without connections, but I believe the issue should be addressed in Hocuspocus.

Perhaps unloading should occur regardless of store document hook result?

**Environment?**
 - Hocuspocus version: 2.13.5

PR #895, opened 2025-01-30
fix: Memory leak in Direct Connection

bencarletonn:
- As described in https://github.com/ueberdosis/hocuspocus/issues/846 there is a scenario where: a server calls .disconnect() on a direct connection to a document, the document is also loaded in an alternative server, occasionally the redis extensions `onStoreDocument` returns a rejected promise due to failing to acquire a lock
- The rejected promise prevents the promise chain `.then` to continue to call `storeDocumentHooks`, preventing `afterStoreDocument`, `unloadDocument` and `afterUnloadDocument` to be called. This is how the memory leak occurs
- The redis extensions `afterStoreDocument` is never triggered. That's fine, we don't need to release a lock that we never acquired.
- More importantly, since we do not trigger the `onDisconnect` hooks when calling .disconnect() on the direct connection, the redis extensions `onDisconnect` logic isn't triggered either. The fix to unload the document associated with the direct connection here https://github.com/ueberdosis/hocuspocus/pull/831/files is never reached.

I don't think we should unload the document if the store document hook returns a rejected promise. Take for example, an autosave that is implemented through a debounced `onStoreDocument`. We wouldn't want the server to suddenly unload the active document if for some reason that autosave failed.

However, if we confirm that the direct connection is the ONLY connection to that document, we can trigger the 'onDisconnect' hook and can call `unloadDocument` for that document when we call .disconnect() on the direct connection. This should ensure the document is unloaded, consequently preventing unused loaded documents from accumulating on the server.

### DIFF
--- a/packages/server/src/DirectConnection.ts
+++ b/packages/server/src/DirectConnection.ts
@@ -74,7 +74,7 @@ export class DirectConnection implements DirectConnectionInterface {
           documentName: this.document.name,
           requestHeaders: {},
           requestParameters: new URLSearchParams(),
-        });
+        })
 
         await this.instance.unloadDocument(this.document);
       }
--- a/packages/server/src/DirectConnection.ts
+++ b/packages/server/src/DirectConnection.ts
@@ -76,7 +76,7 @@ export class DirectConnection implements DirectConnectionInterface {
           requestParameters: new URLSearchParams(),
         })
 
-        await this.instance.unloadDocument(this.document);
+        await this.instance.unloadDocument(this.document)
       }
 
       this.document = null
--- a/packages/server/src/DirectConnection.ts
+++ b/packages/server/src/DirectConnection.ts
@@ -61,6 +61,24 @@ export class DirectConnection implements DirectConnectionInterface {
         socketId: 'server',
       }, true)
 
+      // If the direct connection was the only connection to the document
+      // then we should trigger the onDisconnect hook for
+      // this doc and unload the document
+      if (this.document.getConnectionsCount() === 0) {
+        await this.instance.hooks('onDisconnect', {
+          instance: this.instance,
+          clientsCount: this.document.getConnectionsCount(),
+          context: this.context,
+          document: this.document,
+          socketId: 'server',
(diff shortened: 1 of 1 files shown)


bencarletonn:
@janthurau any update on this one?

bencarletonn:
@janthurau bumping this again since this is causing problems

packages/server/src/DirectConnection.ts
```diff
@@ -61,6 +61,24 @@ export class DirectConnection implements DirectConnectionInterface {
...
+          clientsCount: this.document.getConnectionsCount(),
+          context: this.context,
+          document: this.document,
+          socketId: 'server',
+          documentName: this.document.name,
+          requestHeaders: {},
+          requestParameters: new URLSearchParams(),
+        })
+
+        await this.instance.unloadDocument(this.document)
```
janthurau, line 79: Thank you for raising the issue and your PR! I'm still a bit sceptic to just force-unload the document here.

Maybe it is a good time to refactor the storeDocumentHooks method, so that it actually throws the error. This would allow you to just try again (or ignore it).

janthurau: thinking about it, isnt it enough to just call the onDisconnect hook? The loaded redis extension should take care of the issue then, or is that not working for some reason?

bencarletonn: It is enough to just call the `onDisconnect` hook in this case, since the redis extension has specific logic in it's onDisconnect hook. 

I'm interested to learn why you're skeptical to force-unload the document here. If there is no connections, what instance would we need the document loaded? 

Alternatively I'm happy to remove the forced unloadDocument and just leave the onDisconnect hook. Thanks for getting back to me!

janthurau: As the onDisconnect hook is async, I assume a connecting coming in between running the hook and unloading the document would end up in a broken document. I just noticed that this can already happen in unloadDocument, so I pushed a fix for this (cfbc706f84fc2be216110cabdcd7778fb67ae11a). I think we can merge your PR now :)

PR #900, opened 2025-02-11
fix(server): fast ws disconnect during new conn setup

0xb4lamx:
**Issue**

the WS could be disconnect{ing/ed} during `setUpNewConnection()` which could lead to improper state.

- `connected` hooks will be fired
  -  the followup `disconnected` event will never be fired... _WS graceful close already triggered..._
- onClose callbacks defined outside/after `createConnection()` wont be triggered
  - uncleaned status
- emitting to a closed socket 

**PR includes**

`server/src/ClientConnection.ts`
- WS state check right before WS emit events and trigger of `connected` hooks
- await `connected` hooks

`packages/server/src/Connection.ts`
- fast return if WS is closed

### DIFF
--- a/packages/server/src/ClientConnection.ts
+++ b/packages/server/src/ClientConnection.ts
@@ -163,15 +163,6 @@ export class ClientConnection {
       return this.hooks('beforeHandleMessage', beforeHandleMessagePayload)
     })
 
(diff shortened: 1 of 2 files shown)


janthurau (approved):
Thank you!

Issue #880, opened 2024-11-22
Server handles messages out-of-order when beforeHandleMessage is used


raimohanska:
**Description**

If you use an asynchronous `beforeHandleMessage` hook, you will very likely have messages handled out-of-order. In [here](https://github.com/ueberdosis/hocuspocus/blob/687eef9f91c75becb77acf3fffc7fc39ad986d87/packages/server/src/Connection.ts#L226) there's no mechanism for guaranteeing processing order.

In many cases this is not an issue, but sometimes it leads to inconsistent transient states when processing Y.js updates out of order. I'm not sure what other effects this may have, as this affects all other incoming WebSocket messages as well.

My suggestion would be to guarantee processing order using a queue mechanism.

**Steps to reproduce the bug**

Add an asynchronous `beforeHandleMessage` hook. For worst case, add big random delays in your hook to mess up message processing order totally.

**Expected behavior**

I would expect messages to be processed FIFO.

janthurau:
This will be fixed with #1058

PR #901, opened 2025-02-11
fix(server): message handling order when using `beforeHandleMessage` (#880)

RubaXa:


packages/server/src/Connection.ts
```diff
@@ -216,21 +220,40 @@ export class Connection {
...
+    if (documentName !== this.document.name) {
+      this.nextProcessBeforeHandleMessage()
+      return
+    }
 
     message.writeVarString(documentName)
 
     this.callbacks.beforeHandleMessage(this, data)
       .then(() => {
+        this.nextProcessBeforeHandleMessage()
```
raimohanska, line 247: The ordering of things here seems a bit off to me: you trigger the processing of a next message *before* you've processed this one (below).

raimohanska: The invocation of `this.nextProcessBeforeHandleMessage` will initiate a new async sequence and to me it seems possible that things go out-of-order again.

packages/server/src/Connection.ts
```diff
@@ -41,6 +41,10 @@ export class Connection {
 
   logger: Debugger
 
+  private beforeHandleMessageQueue: Uint8Array[] = []
```
raimohanska, line 44: You might be able to do with just 1 additional state variable if you unshifted the message only after successfully processing it.

packages/server/src/Connection.ts
```diff
@@ -216,21 +220,40 @@ export class Connection {
...
+    if (this.beforeHandleMessageQueue.length > 0 || this.beforeHandleMessageProcessing) {
+      this.beforeHandleMessageQueue.push(data)
+      return
+    }
+
+    this.processBeforeHandleMessage(data)
+  }
+
+  /** Process an incoming message */
+  private processBeforeHandleMessage(data: Uint8Array): void {
```
raimohanska, line 232: The name of the method is out of sync with the comment. I like the comment more, as what this methos does is process an incoming message after applying the beforeHandleMessage hooks.

raimohanska, line 232: I would prefer using async-await syntax for this method. This would allow you to use a `while (this. beforeHandleMessageQueue.length > 0` loop.

PR #904, opened 2025-02-19
Make storeDocumentHooks throw errors instead of silently ignoring them.

janthurau:
Make debounce function asynchronous and improve usage

Refactor the debounce function to return a Promise, enabling better handling in async workflows. Adjust related code to properly await hooks execution and error catching.

Simplify logic in storeDocumentHooks for improved readability and maintainability.

related to #895


janthurau:
nevermind. this breaks the entire hook architecture (as you won't be able to stop the hook chain for a given hook name)

PR #914, opened 2025-03-17
Fix custom HTTP response onRequest

0xb4lamx:
**Issue**
currently, trying [such example](https://tiptap.dev/docs/hocuspocus/server/hooks#on-request) stated in the docs will ends up hard crashing the server with `Error [ERR_HTTP_HEADERS_SENT]: Cannot write headers after they are sent to the client`.

**PR includes**
- adjusting the `requestHandler` to check if the response was already ended before setting the default, based on `response.writableEnded`.

### DIFF
--- a/packages/server/src/Server.ts
+++ b/packages/server/src/Server.ts
@@ -105,8 +105,10 @@ export class Server {
       await this.hocuspocus.hooks('onRequest', { request, response, instance: this.hocuspocus })
 
       // default response if all prior hooks don't interfere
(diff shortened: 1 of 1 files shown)


janthurau:
hm, are you using `return reject();` at the end? That should prevent the mentioned code from running

0xb4lamx:
> hm, are you using `return reject();` at the end? That should prevent the mentioned code from running

ops! it was confusing to reject a promise with 200 response 😵  
missed it at first! it works now! ty

closing this.

PR #917, opened 2025-03-26
refactor: some minor cleanup after v3

nperez0111:
This does a minor refactor to cleanup some things that got added with V3. Namely the package.json runtime import that depended on Node APIs, which could be done at build time instead

### DIFF
--- a/packages/server/src/Server.ts
+++ b/packages/server/src/Server.ts
@@ -3,17 +3,13 @@ import {
   createServer,
 } from 'http'
 import type { ListenOptions } from 'net'
-import { fileURLToPath } from 'node:url'
-import { dirname, join } from 'node:path'
-import fs from 'node:fs'
 import type WebSocket from 'ws'
 import { WebSocketServer  } from 'ws'
 import type { AddressInfo, ServerOptions } from 'ws'
 import kleur from 'kleur'
 import { defaultConfiguration, Hocuspocus } from './Hocuspocus.js'
 import type { Configuration, onListenPayload } from './types'
-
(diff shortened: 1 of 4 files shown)


packages/server/src/Server.ts
```diff
@@ -3,17 +3,13 @@ import {
...
-import fs from 'node:fs'
 import type WebSocket from 'ws'
 import { WebSocketServer  } from 'ws'
 import type { AddressInfo, ServerOptions } from 'ws'
 import kleur from 'kleur'
 import { defaultConfiguration, Hocuspocus } from './Hocuspocus.js'
 import type { Configuration, onListenPayload } from './types'
-
-const meta = JSON.parse(fs.readFileSync(join(dirname(fileURLToPath(import.meta.url)), '../package.json'), 'utf-8'))
+import meta from '../package.json' assert { type: 'json' }
```
nperez0111, line 12: This will pull the package.json at build time, so no need to look for a package.json at runtime.

It even does tree-shaking to just the version that we are using

nperez0111: Hm, so I just removed ts-node (new Node versions can actually run typescript with a flag now: https://nodejs.org/en/learn/typescript/run-natively)

& it seems to be running for me now.

nperez0111:
To fix the tests & to get this to run in ava, I had to rewrite the imports to make sure they follow node's rules for importing (it needs to be the .ts extension & always point to a file, not just a directory).

nperez0111:
Now, I had to get the tests to resolve the source files instead of just referencing the built ones

PR #945, opened 2025-05-19
fix(extension-redis): unsubscribe from pubsub

tomi-bigpi:
Fixes an issue where Redis PubSub is not unsubscribed due to delayed disconnect logic.

The Redis subscription unsubscribe code is behind a `setTimeout` which checks if the `document` instance is still available on the parent `instance`, which for us is never since the document is disposed by the time the `setTimeout` code is called.

Additionally, the existing `unloadDocument` call is suspect, but I left it in with a check for a valid `document` instance.


fuadsaud:
Shouldn't redis disconnect on `afterUnloadDocument`?

Issue #971, opened 2025-07-28
extension-redis can unload document being loaded for new client connection


james-atticus:
**Description**

During a provider client reconnect, we're observing that `before/afterUnloadDocument` hooks are being called twice, after which the Yjs document is destroyed and updates from the client are silently lost.

Logging from the reconnect showing `before/afterUnloadDocument` being called immediately after the document is loaded:

<img width="1498" height="507" alt="Image" src="https://github.com/user-attachments/assets/261a00f3-258e-4999-8048-b804c6aa7b3b" />

Later logs where `Sync/Update` messages are received but there are no corresponding `out` messages to the client nor `Document X changed` log messages:

<img width="1493" height="383" alt="Image" src="https://github.com/user-attachments/assets/ab1f222f-2807-4b0d-8bad-956d7a0489c8" />

**Steps to reproduce the bug**

I haven't been able to reproduce it directly, but I'm pretty sure the bug is caused by the debounced logic in extension-redis:

1. Client disconnects
2. `onDisconnect` hooks are called (and awaited, as of #587)
   a. The Redis hook resolves immediately, even though it starts a timeout to unsub
3. `onClose` is called, document is unloaded because there are no active connections
4. Client reconnects, calling `setUpNewConnection` -> `createDocument`
5. Document is not loaded, so `loadDocument` is called
   a. Document is added to instance with `this.documents.set(documentName, document)` 
   b. `onLoadDocument` hooks are called
6. At some point during the async `onLoadDocument` hooks, the Redis `onDisconnect` timeout triggers
   a. It calls `this.instance.documents.get(documentName)`, which returns the _new_ document being loaded
   b. There are no active connections for the document, so it is unloaded (`connectionCount: 0` in screenshots above)
7. `loadDocument` -> `createDocument` finish running
8. `setUpNewConnection` calls `createConnection` -> `new Connection` -> `this.document.addConnection(this)`
9. Updates come in from client (`connectionCount: 1` in screenshots above) however it's already too late

You can reproduce the broken state by adding an extension:

```
afterLoadDocument: async (data) => {
    data.instance.unloadDocument(data.document);
},
```

**Expected behavior**

extension-redis does not call `unloadDocument` for the newly loaded document. #831 made it so the extension used the latest document from the Hocuspocus instance, rather than the one passed into the hook, which strikes me as odd.

**Environment?**
- Hocuspocus version: 2.15.3

james-atticus:
[...]

PR #972, opened 2025-07-28
fix: delay adding document to instance until connection is about to be established

james-atticus:
Relates to #971.

Only adds the newly-loaded document to `this.documents` once the full load promise has resolved. This should prevent `this.instance.documents.get(documentName);` in extension-redis' `onDisconnect` hook from destroying the document before its associated connection is established.


janthurau:
makes sense! Could you fix the test ("executes the afterUnloadDocument callback when document fails to load")? That hook should not be called when the document fails (not sure why we implemented it to fire in that case, as that feels unexpected)

Issue #997, opened 2025-09-26
If onStoreDocument takes long time, the server may run two onStoreDocument hooks at the same time


raimohanska:
**Description**

I'm wondering if this is a bug, but IMO it would make sense to guarantee that the server doesn't have two concurrent onStoreDocument hooks running at the same time. Now this can occur at least if the onStoreDocument hook takes longer than decounce interval to execute. Also probably when storing immediately on unload.

**Steps to reproduce the bug**

See the test case in commit https://github.com/ueberdosis/hocuspocus/commit/8590a33544ba88e06c0a28cdcb9b317de938cf93

**Expected behavior**

I would expect the server to enqueue store calls so that only one is running that same time. One way to do this could be to check if a store is in progress and if yes, debounce a new store instead of storing immediately.

**Additional context**

Discovered this while working on #995.

janthurau:
Thank you! I have actually noticed this issue when working on something else, but havent had time to look into it so far. I will try to check your fix out next week!

janthurau:
Thank you for your contribution 🙏

raimohanska:
No problem, was a pleasure to contribute to something I find valuable. The contribution documentation and the great test suite made it quite comfortable too.

PR #998, opened 2025-09-26
Fix two onStoreDocument hooks running that the same time

raimohanska:
Fix for #997. 

I still suspect that adding DirectConnections to the mix will lead to inconsistent behavior but have no test cases for that yet. In any case I believe this fix will make things better.


raimohanska:
Update: I applied async-mutex instead of using the `saving` counter and this approach seems to work with the additional test cases, ensuring that two saves are not executed concurrently.

raimohanska:
Before merge, we might clean up the commit history.

Issue #913, opened 2025-03-14
Hocuspocus v3


janthurau:

julianpoy:
The subdocument ticket (https://github.com/ueberdosis/hocuspocus/issues/583) has the "next-major" tag on it. Will that be part of Hocuspocus v3?

janthurau:
@julianpoy yjs-like subdocument support won't be included in v3, but you can utilize multiplexing and just split documents like documentId/main ; documentId/sub1 ; documentId/sub2, which gives you the same set of benefits.

Palanikannan1437:
The issue with Multiplexing right now is that when the provider is destroyed (irrespective of what options are passed, the websocket connection is closed), even if I destroy it after a delay or something as @janthurau suggested in of the comments. Can we fix this?

branks:
Can you confirm what `preserveConnection` does in v3 please? It seems to only exist in the readme but none of the code?

fuadsaud:
Is there an estimated release date for v3? We see a lot of bug fixes are getting bundled with that release and are curious when we could start testing it.

Palanikannan1437:
Hey @janthurau, in the `HocuspocusProvider`, I'm not able to pass parameters anymore, could you please tell me how to fix this?

Diagnostics:
1. Object literal may only specify known properties, and 'parameters' does not exist in type 'HocuspocusProviderConfiguration'. [2353]

Palanikannan1437:
Hey @janthurau, there's an issue with the example and maybe the underlying code as well, the connection status remains in a "connecting" status until you've manually clicked on "connect" that calls `provider.configuration.websocketProvider.connect()` which then closes previous connections that are pending and opens a new one anyways.

Is this a bug?

Palanikannan1437:
Hey @janthurau, there's an issue with the example and maybe the underlying code as well, the connection status remains in a "connecting" status until you've manually clicked on "connect" that calls `provider.configuration.websocketProvider.connect()` which then closes previous connections that are pending and opens a new one anyways.

Is this a bug?

And also is there any way I could create the websocket connection outside of HocusPocus and pass that connection to it?

tehbelinda:
Came here to look for answers on what happened to `parameters` and that was helpful, thanks! 

@janthurau Is it possible to add an upgrade guide for HocuspocusProvider too for changes like this? Or perhaps add it to the release notes somewhere? Would be really helpful for those of us trying to figure out the impact of the upgrade 🙏

PR #1006, opened 2025-10-07
docs: mention requirement to call attach() when passing websocketProvider instead of url

roschlau:
As mentioned in this issue, but apparently not documented anywhere else yet: https://github.com/ueberdosis/hocuspocus/issues/913#issuecomment-2815270137

Spent way too much time trying to figure out why our hocuspocus provider just wasn't doing anything anymore after upgrading to 3.1, hoping to spare others with this change. I put this in the place where it seemed most logical to me, feel free to let me know if it should also be mentioned somewhere else or moved.


janthurau:
Thank you! This was added in the new repo (tiptap-docs).

PR #1059, opened 2026-02-06
feat: first draft of @hocuspocus/provider-react package

janthurau:
The idea of this package is to remove all issues around rerendering during development in React. Strict mode triggers hooks twice, which is confusing (the network tab shows two websockets, one is dead and another is connected), and can cause timing issues (the server doesnt support opening the same document name twice over the same websocket connection), so if a disconnect and connect happen really fast, the state may mix and connections break.

Status: Needs careful testing.


packages/provider-react/src/HocuspocusProviderWebsocketComponent.tsx
```diff
@@ -0,0 +1,75 @@
...
+ *   </HocuspocusRoom>
+ * </HocuspocusProviderWebsocketComponent>
+ * ```
+ */
+export function HocuspocusProviderWebsocketComponent({
+	children,
+	url,
+	websocketProvider: externalWebsocketProvider,
+}: HocuspocusProviderWebsocketComponentProps) {
+	const websocketRef = useRef<HocuspocusProviderWebsocket | null>(null);
```
bdbch, line 29: One thing I would look out here is React's strict mode where this could lead to two hocuspocus instances being created. Not sure if its a problem - just wanted to point it out.

packages/provider-react/src/HocuspocusRoom.tsx
```diff
@@ -0,0 +1,156 @@
...
+					document,
+					// token is intentionally read from the current closure but not
+					// included as a dependency — token refreshes should not destroy
+					// the connection. Function tokens are called on-demand by the provider.
+					token,
+				}),
+			);
+		}
+		// eslint-disable-next-line react-hooks/exhaustive-deps
+	}, [name, document, websocketProvider]);
```
bdbch: Should the token also be a dependency here? Otherwise stale tokens from the first render would be used.

bdbch (changes requested):
small things

PR #1096, opened 2026-05-18
feat(server): add beforeHandleAwareness hook

### LINKED ISSUE #1095: Add a hook to modify awareness updates before broadcast
**The problem I am facing**

I need to modify an incoming awareness update on the server before it is broadcast to peers. The motivating case is stamping a server-validated user identity (from `onAuthenticate` / `connection.context.user`) onto `state.user` so cursor labels cannot be spoofed by clients.

The closest existing hook is `onAwarenessUpdate`, but it runs after `Document.handleAwarenessUpdate` has already broadcast the original payload. `Document.handleAwarenessUpdate` is attached in the `Document` constructor and ends up first in the listener chain; `onAwarenessUpdate` is wired later, in `loadDocument`. Mutations the hook makes therefore only reach peers in a second frame moments later. The CRDT converges on the corrected value, but the original payload still leaves the server and is visible to DevTools and any custom y-awareness listener on the client.

**The solution I would like**

Either of the following would work:

1. A new `beforeAwarenessUpdate` hook fired before `Document.handleAwarenessUpdate`, where the hook can mutate or reject the awareness state and the broadcast picks up its changes. This mirrors `beforeHandleMessage` but at the awareness layer.
2. Reorder the listener chain so `onAwarenessUpdate` runs before the broadcast, and document that mutations made from inside it are included in the outgoing frame.

The contract in either case: whatever a hook writes into `awareness.getStates().get(clientId)` is what peers receive.

- `beforeHandleMessage` with byte-level rewriting. Works, but requires decoding the y-protocols wire format and tracking future protocol changes.
- Living with the double-broadcast and relying on CRDT convergence. Browsers usually only paint the corrected state, but the original payload is still on the wire and observable.
- Client-side identity stamping. Moves the trust boundary back to the client, which defeats the goal.

**Additional context**

Repro (assumes #1094 is merged, otherwise `payload.connection` is `undefined`):

```ts
new Server({
  async onAuthenticate({ token }) {
    return { user: { id: 'authoritative', name: 'Authoritative' } }
  },
  async onAwarenessUpdate({ awareness, connection, added, updated }) {
    const expected = connection?.context?.user
    if (!expected) return
    for (const clientId of [...added, ...updated]) {
      const state = awareness.getStates().get(clientId)
      if (!state) continue
      state.user = expected
      awareness.emit('update', [{ added: [], updated: [clientId], removed: [] }, 'server'])
    }
  },
})
```

Connect a client, call `awareness.setLocalStateField('user', { name: 'Spoofed' })`. In a peer's DevTools `Network → WS`, two consecutive frames appear: spoofed, then corrected.

I am happy to put up a PR for whichever shape fits the project. Asking first because the answer likely touches the public hook contract.

Related: #1094.

dschmidt:
Adds a new hook that fires after decoding an inbound awareness message and before `applyAwarenessUpdate` runs. The payload exposes the per-client states as a mutable `Map<clientId, state>`. Mutations rewrite the encoded update that is then applied and broadcast; throwing rejects the update.

Use cases include server-side identity stamping (anti-spoofing), state validation, and filtering.

Payload mirrors `onAwarenessUpdatePayload`: `transactionOrigin: unknown` plus `connection?` convenience shortcut, derived via `isTransactionOrigin`. Extensions chain naturally and run before the configuration-level hook.

Follow-up to #1094 (which fixed the awareness path to wrap origin as `TransactionOrigin`).

Fixes #1095

### DIFF
--- a/packages/server/src/Document.ts
+++ b/packages/server/src/Document.ts
@@ -21,6 +21,11 @@ export class Document extends Doc {
 		// eslint-disable-next-line @typescript-eslint/no-empty-function
 		onUpdate: (document: Document, origin: unknown, update: Uint8Array) => {},
 		beforeBroadcastStateless: (document: Document, stateless: string) => {},
+		beforeHandleAwareness: (
+			document: Document,
+			states: Map<number, Record<string, any>>,
+			transactionOrigin: unknown,
+		) => Promise.resolve(),
 	};
 
 	connections: Map<
@@ -96,6 +101,27 @@ export class Document extends Doc {
 		callback: (document: Document, stateless: string) => void,
 	): Document {
 		this.callbacks.beforeBroadcastStateless = callback;
+
+		return this;
+	}
+
+	/**
+	 * Set a callback that will be triggered before an inbound awareness update
+	 * is applied to this document's awareness state. The callback receives the
+	 * document, the decoded per-client states as a mutable `Map`, and the
+	 * `TransactionOrigin` that will be forwarded to `applyAwarenessUpdate`.
+	 * Use `isTransactionOrigin(origin)` to discriminate sources. Mutate the
+	 * map in place (set/delete/field changes) to rewrite the update, or throw
+	 * to reject it entirely.
+	 */
+	beforeHandleAwareness(
+		callback: (
+			document: Document,
+			states: Map<number, Record<string, any>>,
+			transactionOrigin: unknown,
+		) => Promise<any>,
+	): Document {
+		this.callbacks.beforeHandleAwareness = callback;
 
 		return this;
 	}
--- a/packages/server/src/Hocuspocus.ts
+++ b/packages/server/src/Hocuspocus.ts
@@ -46,6 +46,7 @@ export class Hocuspocus<Context = any> {
 		onConnect: () => new Promise((r) => r(null)),
 		connected: () => new Promise((r) => r(null)),
 		beforeHandleMessage: () => new Promise((r) => r(null)),
+		beforeHandleAwareness: () => new Promise<void>((r) => r()),
 		beforeSync: () => new Promise((r) => r(null)),
 		beforeBroadcastStateless: () => new Promise((r) => r(null)),
 		onStateless: () => new Promise((r) => r(null)),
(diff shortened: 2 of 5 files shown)


dschmidt:
**Real-world usage in OpenCloud's collaboration sidecar PoC**

I'm using this hook in a Yjs/Hocuspocus-based realtime collaboration PoC for OpenCloud to enforce server-side user identity. The threat model: a malicious client could send `awareness.user = { id: 'admin', name: 'Admin' }` and peers would render that as truth. With `beforeHandleAwareness` I overwrite the `user` field with the authenticated identity from `connection.context` before any broadcast happens.

The full implementation is ~15 lines:

```js
new Server({
  async onAuthenticate({ token, requestParameters }) {
    const me = await validateTokenAgainstOpenCloud(token)
    const access = await probeFileAccess(token, requestParameters.get('fileId'))
    if (access === null) throw new Error('access denied')
    return { readOnly: !access.canWrite, user: { id: me.id, displayName: me.displayName, color: deterministicColor(me.id) } }
  },

  async beforeHandleAwareness({ states, connection }) {
    const user = connection?.context?.user
    if (!user) return
    const canonical = { id: user.id, name: user.displayName, color: user.color }
    for (const state of states.values()) {
      state.user = canonical
    }
  }
})
```

**What I had before** (workaround on a Hocuspocus v4.0.0 instance):

```js
async afterLoadDocument({ document }) {
  document.awareness.on('update', (changes, origin) => {
    // stamp identity post-hoc + awareness.emit('update', ..., 'server')
  })
}
```

This had two problems the new hook removes:

1. **Broadcast race:** `applyAwarenessUpdate` emits `update` *before* the listener runs, so the spoofed state was already on the wire for one tick before the corrective re-emit.
2. **Re-encoding cost:** every correction triggered a second broadcast.

With `beforeHandleAwareness` it's a single pass: decode → hook mutates the states map → re-encode → apply → broadcast. The mutate-in-place API also chains cleanly when multiple extensions need to inspect or rewrite states.

janthurau (approved):

dschmidt:
Nice, that was quick! Thanks a lot! :)

PR #1123, opened 2026-07-14
fix: when beforeHandleMessage throws, we don't want to process other messages that were already queued

janthurau:
If you reject a message in beforeHandleMessage (e.g. during validation), the client is probably doing something weird and you don't want following messages (that were queued before the connection got closed) to be processed.


janthurau:
@coderabbitai review

PR #1130, opened 2026-07-28
Feat/batch updates before sending to clients

janthurau:
When a lot of users connect to the same document really fast after the other, or all of them are emitting changes at around the same time (reproduced with 400+ users on a single doc that apply changes), the server may end up with a blocked main loop because every change is sent to every user. So if you have 400 users changing the doc, the server loops through 400 changes and each time loops through 400 users to emit the update, easily ending up with a few hundred thousand messsages sent).

This feature allows the server to merge updates if they arrive within `flushDelay` and thus significantly reduce the number of messages sent, while only adding a slight delay.

The default `flushDelay` is just a `setImmediate`. It can be set to any number in milliseconds (also `0` which is the default), or `false` to turn it off.

### DIFF
--- a/packages/server/src/Document.ts
+++ b/packages/server/src/Document.ts
@@ -10,6 +10,7 @@ import { OutgoingMessage } from "./OutgoingMessage.ts";
 import type {
 	AwarenessUpdate,
 	ConnectionTransactionOrigin,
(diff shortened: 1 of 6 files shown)


janthurau:
@coderabbitai review

